PR #3, opened 2018-02-06
Add config for ResultStore

willsc79:


xingao267:
Do we want to update https://github.com/bazelbuild/bazel-toolchains/blob/master/bazelrc/bazel-0.10.0.bazelrc as well, or this is only for bazel 0.9.0?

bergsieker:
Yes, we should update the 0.10.0 configs, too.

willsc79:
I'll do this. Also the email address for the commits got messed up, I'm going to send another pull request...

PR #4, opened 2018-02-07
Add BES flags

willsc79:


willsc79:
I added the text that Michelle suggested.

PR #6, opened 2018-02-28
add rbe-debian8 container by container_layer

erain:
* make use of container_layer rule to generate the experimental
rbe-debian8 container.
* rbe-debian8 container can pass the same tests as the current
debian8-clang-fully-loaded container
* since `layers` attr is used in container_image rule, we rename
`layers` -> `language_layers`

Tested:
- bazel test //container/debian8-clang-fully-loaded:all
- bazel test //container/experimental/rbe-debian8:all

Change-Id: I7c4a9e626d720d94f6f730f31f3f4d57c35a9a10

### DIFF
--- a/container/experimental/rbe-debian8/BUILD
+++ b/container/experimental/rbe-debian8/BUILD
@@ -19,11 +19,6 @@ package(default_visibility = ["//visibility:public"])
 load("@distroless//cacerts:cacerts.bzl", "cacerts")
 load("@jessie_package_bundle//file:packages.bzl", "packages")
 load(
-    "//container/rules:docker_toolchains.bzl",
-    "language_tool_layer",
-    "toolchain_container",
-)
-load(
     "//skylib:packages.bzl",
     "base_layer_packages",
     "clang_layer_packages",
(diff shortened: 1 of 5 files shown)


container/test/rbe-debian8.yaml
```diff
@@ -1,9 +1,21 @@
 schemaVersion: "1.0.0"
 
 commandTests:
```
nlopezgi: please add a couple more tests to verify that the base packages are installed properly (e.g., check git, wget, etc)

container/experimental/rbe-debian8/BUILD
```diff
@@ -0,0 +1,113 @@
...
+        "base-ltl",
+        "clang-ltl",
+        "go-ltl",
+        "java-ltl",
+        "python-ltl",
+    ],
+    tags = ["manual"],
+)
+
+container_layer(
```
nlopezgi: as discussed offline, we should make changes upstream so that this can be a container_image target as well, but you can submit with this as it is.

container/test/rbe-debian8.yaml
erain: Done. Tests added to verify all language interpreters/compilers and tools.

container/experimental/rbe-debian8/BUILD
erain: Implement compossible `container_image` in [rules_docker 335](https://github.com/bazelbuild/rules_docker/pull/335). 

As discussed, leave `container_layer`s as they are right now, and will change to `container_image` if my PR got merged.

erain:
The actual tests passed, but task failed itself during cleanup: 
https://buildkite.com/bazel/bazel-toolchains/builds/65#5b3880dc-358f-417d-affb-d2ab137fb4e0

Nothing is really broken, so I will merge this PR. 

@philwo Could you help take a look at the test and let us know about how to fix it? Many Thanks!

PR #11, opened 2018-03-09
Migrate to new package manager.

xingao267:
Change-Id: I82c65a7734eb0118b8a95309434eaded86d5c284

Previously reviewed at https://bazel-review.googlesource.com/#/c/bazel-toolchains/+/39650/


xingao267:
@philwo I modified the location of tests target and updated the presubmit.yaml file, but the new yaml file doesn't seem to be picked up by Buildkite. Do you maybe have an idea why?

xingao267:
It seems that it is using --http_config=https://raw.githubusercontent.com/bazelbuild/continuous-integration/master/buildkite/pipelines/bazel-toolchains-postsubmit.yml. Should Buildkite look at https://github.com/bazelbuild/bazel-toolchains/blob/master/.bazelci/presubmit.yml instead?

buchgr:
@xingao267 yes it should use the checked in yaml. I have updated the pipeline. Can you please try again?

xingao267:
@buchgr Thanks! The presubmit is picking up the checked-in yml file. Does the postsubmit also use the same checked-in file?

buchgr:
@xingao267 yes it does.

xingao267:
@buchgr Great! Thanks!

PR #16, opened 2018-03-21
Add --all_incompatible_changes flag when generating configs.

xingao267:


rules/docker_config.bzl
```diff
@@ -259,7 +259,8 @@ def _docker_toolchain_autoconfig_impl(ctx):
     bazel_cmd += " && touch WORKSPACE && mv BUILD.sample BUILD"
   # For each config repo we run the target @<config_repo>//...
   bazel_targets = "@" + "//... @".join(ctx.attr.config_repos) + "//..."
-  bazel_cmd += " && bazel build " + bazel_targets
+  bazel_flags = "--all_incompatible_changes"
```
erain: Could you add spaces around `bazel_flags`? It is easier to use `+=` operator in the future if we need more flags

xingao267: Done.

PR #27, opened 2018-04-11
Add experimental platform and toolchain definitions.

xingao267:
- experimental rbe_debian8 platform and toolchain definitions
- experimental ubuntu_gcc_fully_loaded platform and toolchain
definitions


platform/experimental/BUILD
```diff
@@ -0,0 +1,49 @@
+package(default_visibility = ["//visibility:public"])
```
nlopezgi: rename directory 'platforms'

nlopezgi: set license headers

platform/experimental/BUILD
```diff
@@ -0,0 +1,49 @@
...
+        "@bazel_tools//tools/cpp:gcc",
+        ":trusty",
+    ],
+    remote_execution_properties = """
+          properties: {
+            name: "container-image"
+            value:"docker://gcr.io/asci-toolchain/nosla-ubuntu-gcc-fl@sha256:20bf8513ba378ff3288904ab73b9993586c731e42855e27cf3be42c785a42489"
+         }
+         """,
+)
```
nlopezgi: Lets only add the debian8 for now and only migrate the debian8 targets

platform/experimental/BUILD
xingao267: done.
Also rename toolchain to toolchains if that's fine with you

platform/experimental/BUILD
xingao267: done

PR #31, opened 2018-04-13
Add feature to specify tarball of debs through a flag.

xingao267:
Downloadble link of tarball of debs and sha256 can be specified
through --action_env=DOWNLOADABLE_TAR=<link> and
--action_env=DOWNLOADABLE_TAR_SHA256=<sha256>.


container/rules/docker_toolchains.bzl
```diff
@@ -62,16 +63,43 @@ def _language_tool_layer_impl(ctx, symlinks=None, env=None, tars=None,
...
+  if "DOWNLOADABLE_TAR" in ctx.configuration.default_shell_env:
+    if "DOWNLOADABLE_TAR_SHA256" not in ctx.configuration.default_shell_env:
+      fail("'DOWNLOADABLE_TAR_SHA256' must be specified when 'DOWNLOADABLE_TAR' is specified.")
+    installables_tar = ctx.actions.declare_file(ctx.attr.name + "-packages.tar")
+
+    wget_command = """\
+#!/bin/bash
+set -e
+# Execute 'wget' command inside a docker container in case the host does not have 'wget' installed.
+cid=$(docker run -w="/" -e DOWNLOADABLE_TAR -d --privileged launcher.gcr.io/google/rbe-debian8:latest sh -c "wget $DOWNLOADABLE_TAR --no-verbose -O /tmp/debs.tar")
```
xingao267: I'm a little unsure about reusing rbe-debian8 container here. The reason why I have to do wget in a container is because the bazel container on GCB doesn't have wget installed. So I just picked a container that have wget in it, and the first one come to my mind is our container. Let me know if you have any ideas.

container/rules/docker_toolchains.bzl
```diff
@@ -20,20 +20,22 @@ load("@debian_docker//package_managers:install_pkgs.bzl", _install = "install")
...
 
   has_no_packages = "packages" not in kwargs or kwargs["packages"] == []
   if has_no_packages and "additional_repos" in kwargs:
     fail("'additional_repos' can only be specified when 'packages' is not empty.")
 
-  has_no_tar = "installables_tar" not in kwargs or kwargs["installables_tar"] == ""
-  if has_no_packages and has_no_tar and "installation_cleanup_commands" in kwargs:
-    fail("'installation_cleanup_commands' can only be specified when at least " +
-      "one of 'packages' or 'installables_tar' is not empty.")
+  if has_no_packages and "installation_cleanup_commands" in kwargs:
```
nlopezgi: does installation via tar / action_env not support cleanup commands?

container/rules/docker_toolchains.bzl
```diff
@@ -62,16 +63,43 @@ def _language_tool_layer_impl(ctx, symlinks=None, env=None, tars=None,
...
-  installables_tars = installables_tars or []
-  if installables_tars == [] and ctx.file.installables_tar:
-    installables_tars = [ctx.file.installables_tar]
   installation_cleanup_commands = installation_cleanup_commands or ctx.attr.installation_cleanup_commands
+  installables_tar = None
+
+  # If a link to the tarball of deb packages is passed in through
+  # --action_env=DOWNLOADABLE_TAR=<link>, use that one.
+  if "DOWNLOADABLE_TAR" in ctx.configuration.default_shell_env:
+    if "DOWNLOADABLE_TAR_SHA256" not in ctx.configuration.default_shell_env:
```
nlopezgi: this should be in the input validation

nlopezgi: My concern is that we will only be able to pass one single tar/sha via the flag (so if we wanted to build multiple layers in the same command this could be limiting)
Lets try to do it this way:
have the sha be defined in the BUILD rule, along with a name of the tar file, and only pass via the flag the gcs bucket?

container/rules/docker_toolchains.bzl
nlopezgi: don't like this, we instead need to have a container on GCB that does have wget installed and make wget part of the required tools for our container rules. That also means we will need a custom container for GCB that has bazel + wget. I'm pretty sure we have done this with Yu in the past

erain: Yes, to use a custom docker container with GCB you can reference [Creating Custom Build Steps](https://cloud.google.com/container-builder/docs/create-custom-build-steps). I can help you with that.

container/rules/docker_toolchains.bzl
xingao267: action_envs seems can't be accessed in macro level (i.e. from kwargs).

xingao267:
Discussed offline internally. We will no longer add this flag feature.

PR #39, opened 2018-04-24
Update README.md

davidstanke:
In my testing, the 0.10.0 config also works with Bazel 0.11.0 and 0.12.0. Do you feel confident that it will work across use cases? If so, let's add this text so users will know they can keep using this config.


xingao267:
PR 38 is just submitted.

PR #42, opened 2018-04-30
adding libcxx (msan version) to all rbe containers

nlopezgi:
* Also moved reference to latest version of clang (to match smallest one with libcxx)


container/experimental/rbe-debian9/BUILD
```diff
@@ -74,11 +74,15 @@ language_tool_layer(
     env = {
```
xingao267: We should probably update experimental/rbe-debian8 as well

third_party/clang/revision.bzl
```diff
@@ -1,4 +1,4 @@
-CLANG_REVISION="r327695"
-DEBIAN8_CLANG_SHA256="588ddbd83c32e7ab87fad58038c3712f6d1b0056f12ea746169ac7ecd33c93bf"
-DEBIAN9_CLANG_SHA256="434bbe2195845e42eb36ad2b88f6aec658d192614d47f71e15b861e09f71b4bd"
-UBUNTU16_04_CLANG_SHA256="48bfc470810d483ffb2e826ed3540efc1da45fa89f416f8efa439242495e6239"
+CLANG_REVISION="r328903"
```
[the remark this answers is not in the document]
nlopezgi: I think the next container release might include releasing the other containers, so we do want to do a release but not in this cl.

container/experimental/rbe-debian9/BUILD
nlopezgi: done

nlopezgi:
verified the failing targets are now passing in my local env. Errror was probably related to use of outdated rc version in those tests.

PR #46, opened 2018-05-04
Fix python setups for rbe-{debian8, debian9, ubuntu16_04}

erain:
This PR is aimed for fixing various issues in python setup in all three images, so that later both py2/py3's pip package manager would work better.


erain:
Sure, I will test rbe-{debian8, ubuntu16_04} with some builds before submitting this PR.

PR #50, opened 2018-05-11
container_install_pkgs supports chaining installation of packages

nlopezgi:
- Previously rule could not be used if base was also produced by container_install_pkgs.
- Also added some interesting examples/tests of how packages can be installed.


rules/docker_config.bzl
```diff
@@ -152,7 +152,7 @@ tar_filetype = [
     ".tar.xz",
 ]
 
-def container_install_pkgs(name, base, packages, additional_repos, keys):
+def container_install_pkgs(name, base, packages, additional_repos = [], keys = [], tags = []):
```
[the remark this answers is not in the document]
nlopezgi: According to the thread it states that behavior is currently that it does apply for build. Tried running bazel build ... and indeed the target is not executed. Let me know if you'd still rather I remove it.

test/container/BUILD
```diff
@@ -15,3 +15,107 @@
...
+      packages = [
+        "docker-ce",
+      ],
+)
+
+install_pkgs(
+      name = "docker",
+      image_tar = "@debian8//image",
+      installables_tar = ":test-docker-packages.tar",
+      output_image_name = "bazel/container/rbe-debian8:docker",
```
xingao267: what about bazel/container/docker-debian8

It's not really based on rbe-debian8

nlopezgi: fixed (copy paste error)

PR #56, opened 2018-05-17
updating experimental toolchain rule to use 0.13.0 cc_toolchain

nlopezgi:


xingao267:
Do we want to update the platform to the latest container as well?

nlopezgi:
It already is pointing to the latest container: https://github.com/bazelbuild/bazel-toolchains/commit/aeaf869ca063806f9352bc1fd1cd1b8a34473631

xingao267:
I see

PR #59, opened 2018-05-18
Move platform and toolchain definition out of experimental.

xingao267:


constraints/BUILD
```diff
@@ -24,19 +24,14 @@ constraint_value(
...
-            name: "container-image"
-            value:"docker://gcr.io/cloud-marketplace/google/rbe-debian8@sha256:d84a7de5175a22505209f56b02f1da20ccec64880da09ee38eaef3670fbd2a56"
-         }
-         """,
+constraint_value(
+    name = "xenial",
+    constraint_setting = ":linux_distribution",
+)
+
+constraint_setting(name = "msan")
```
nlopezgi: name = sanitizer

configs/debian8_clang/0.3.0/bazel_0.13.0/cpp/BUILD
```diff
@@ -30,3 +30,19 @@ toolchain(
...
     toolchain_type = "@bazel_tools//tools/cpp:toolchain_type",
 )
+
+toolchain(
+    name = "cc-toolchain-clang-x86_64-msan",
+    exec_compatible_with = [
+        "@bazel_tools//platforms:linux",
+        "@bazel_tools//platforms:x86_64",
+        "@bazel_tools//tools/cpp:clang",
+        "//constraints:support_msan",
```
nlopezgi: should this list include jessie?

configs/debian8_clang/0.3.0/BUILD
```diff
@@ -24,3 +24,28 @@ java_runtime(
...
+)
+
+# RBE Debian8 r328903
+platform(
+    name = "rbe_debian8_r328903",
+    constraint_values = [
+        "@bazel_tools//platforms:x86_64",
+        "@bazel_tools//platforms:linux",
+        "@bazel_tools//tools/cpp:clang",
+        "//constraints:support_msan",
```
nlopezgi: just to double check, resolution is such that the toolchain's exec_compatible_with can be a subset of the values here (i.e., the non msan toolchain also satisfies this platform)?

configs/debian8_clang/0.3.0/bazel_0.13.0/cpp/BUILD
nlopezgi: and also for the toolchain above?

configs/ubuntu16_04_clang/1.0/bazel_0.13.0/cpp/BUILD
```diff
@@ -0,0 +1,49 @@
...
+licenses(["notice"])  # Apache 2.0
+
+package(default_visibility = ["//visibility:public"])
+
+toolchain(
+    name = "cc-toolchain-clang-x86_64-default",
+    exec_compatible_with = [
+        "@bazel_tools//platforms:linux",
+        "@bazel_tools//platforms:x86_64",
+        "@bazel_tools//tools/cpp:clang",
```
nlopezgi: should this list include xenial?

constraints/BUILD
```diff
@@ -24,19 +24,14 @@ constraint_value(
...
+
+constraint_value(
+    name = "support_msan",
```
nlopezgi: just "msan"

constraints/BUILD
nlopezgi: or sanitizer_support

xingao267: I tried that. I had this before

constraint_setting(name = "sanitizer")

constraint_value(
    name = "no_sanitizer",
    constraint_setting = ":sanitizer",
)

constraint_value(
    name = "msan",
    constraint_setting = ":sanitizer",
)

But it doesn't work, for one platform, it can only have one value for one constraint_settings. So if a platform supports msan, tsan, no_sanitizer, I couldn't do it unless create duplicate platforms, each has one sanitizer constraint value, and with the same container sha.

We sort of have to do it in the support_msan, support_tsan, way.

constraints/BUILD
xingao267: see above.

constraints/BUILD
nlopezgi: got it, lgtm as it is then

nlopezgi: Oh, I think there's a better way though. In constraints create folder sanitizers with a build file there, call the constraint just msan

PR #61, opened 2018-05-24
update container docs to point to ubuntu16-04

nlopezgi:
* debian8 is being deprecated, point to ubuntu16-04 in docs


buchgr:
@erain yes I created new images because our ubuntu was broken for some reason.

PR #62, opened 2018-05-25
update hash of bazel's gpg key

buchgr:
Bazel's gpg key expired yesterday and so we updated the expiration date, leading to a different hash of the public key.

This currently breaks the build

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -107,7 +107,7 @@ http_archive(
 
 http_file(
     name = "bazel_gpg",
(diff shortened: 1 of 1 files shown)


nlopezgi:
There's another test that is failing. I'm including a fix for both the gpg issue and the other failing test in PR https://github.com/bazelbuild/bazel-toolchains/pull/61

nlopezgi:
closing as #61 includes this change as well.

PR #64, opened 2018-05-30
replace ctx.actions.run for run_shell

nlopezgi:
* either sh_binary or ctx.actions.run are not compatible with remote
execution. run_shell is.

### DIFF
--- a/skylib/BUILD
+++ b/skylib/BUILD
@@ -16,7 +16,4 @@ licenses(["notice"])  # Apache 2.0
 
 package(default_visibility = ["//visibility:public"])
 
-sh_binary(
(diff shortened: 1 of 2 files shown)


nlopezgi:
sent this just as an fyi for now (and to verify it passes checks),  but I'll try some other alternatives as im still not 100% sure that the cuplrit target is sh_binary or ctx.actions.run

nlopezgi:
Ok, looked in detail at sh_binary and it just outputs a symlink to the file, so I'm almost sure ctx.actions.run is the issue. This is now ready for review.

erain:
It looks like this PR breaks the CI, retry the test.

PR #65, opened 2018-05-31
Fix tar concatenation.

xingao267:


container/rules/docker_toolchains.bzl
```diff
@@ -139,7 +139,7 @@ def _language_tool_layer_impl(ctx, symlinks=None, env=None, tars=None,
     ctx.actions.run_shell(
       inputs=installables_tars,
       outputs=[final_installables_tar],
-      command="tar cvf {output_tar} --files-from /dev/null && tar --concatenate -v --file={output_tar} {input_tars}".format(
+      command="tar cvf {output_tar} --files-from /dev/null && for i in {input_tars}; do tar --concatenate --file={output_tar} $i; done".format(
```
nlopezgi: Can we wrap this line. Also, could you comment on the PR comment what was the issue with how this was done before?

xingao267:
It seems that concatenate multiple tarballs in one tar --concatenate command doesn't work all the time. I experienced the same error similar to this one (https://www.queryxchange.com/q/2_941475/concatenate-multiple-tar-files-in-one-command/). It was working before, but didn't work anymore. I don't exactly why it starts failing. I suspect it could be because we switch from Debian8 to Ubuntu16_04. It is safer to just concatenate them one by one.

container/rules/docker_toolchains.bzl
xingao267: Done. Added comments as well.

PR #71, opened 2018-06-06
fix default cmd (bash) and locale (C.UTF-8)

erain:
Tested: rbe-{debian8,debian9,ubuntu16_04} are all tested with Google
Container Builder


erain:
Note on using `cmd` over `entrypoint`: this is the base images (`launcher.gcr.io/google/debian8` etc.) are using `cmd`.

PR #75, opened 2018-06-08
revert clang r331746 to r328903

erain:
* clang r331746 failed on some of the tests and may not be stable.
  Revert to the last stable revision r328903


erain:
We didn't revert #72 directly, because there were python version update as well as unit tests for libcxx installation. Also there were some move around of test targets in other PRs.

PR #76, opened 2018-06-08
Organize LTLs and yamls to layer folder structure.

xingao267:


container/common/yamls/BUILD
```diff
@@ -0,0 +1,19 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
nlopezgi: All of the yaml files in this directory are for tests / validations of the containers, right? Not sure I like the name 'yamls' if that is the case. Sounds more like something that should be in a test directory (or just under common, if we have nothing else there)

xingao267: Moved them to directly under common/

PR #77, opened 2018-06-08
bazelrc: set bes timeout to 60 seconds and add BES URL flag to 0.14.0

willsc79:


bazelrc/bazel-0.14.0.bazelrc
```diff
@@ -0,0 +1,89 @@
+# Copyright 2016 The Bazel Authors. All rights reserved.
```
nlopezgi: We will be releasing configs for 0.14.0 shortly. You can send your change for all the previous versions for now. Please let us know which new flags will be needed for BES for 0.14.0 so that we can include them in that release

willsc79: ok cool. The new flags I need are:

build:results --bes_results_url="https://source.cloud.google.com/results/invocations/"

build:results-local --bes_results_url="https://source.cloud.google.com/results/invocations/"

willsc79: also setting 

build:results --bes_timeout=60s

build:results-local --bes_timeout=60s

would be good.

willsc79: thanks! I'm happy to review the changes if you want.

PR #84, opened 2018-06-12
Adding configs for ubsan for bazel 0.14.1

nlopezgi:


nlopezgi:
No, we do not want to include ubsan in the toolchain.bazelrc as it's still experimental

PR #89, opened 2018-06-18
Update documents.

xingao267:


xingao267:
There is another race condition caused by install_pkgs rule as image all the images from container_pull have the tag bazel/image:image. Arthur is working on a fix in base-image-docker.

PR #93, opened 2018-06-19
Add wget to bazel container.

xingao267:
- remove some out-of-date comments.
- fix package tarball name

### DIFF
--- a/container/ubuntu16_04/builders/bazel/BUILD
+++ b/container/ubuntu16_04/builders/bazel/BUILD
@@ -38,6 +38,7 @@ toolchain_container(
         "//container/ubuntu16_04/layers/bazel:bazel-tools",
         "//container/ubuntu16_04/layers/bazel:bazel-ltl",
         "//container/ubuntu16_04/layers/git:git-ltl",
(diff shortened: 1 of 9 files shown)


container/ubuntu16_04/builders/bazel/BUILD
```diff
@@ -38,6 +38,7 @@ toolchain_container(
         "//container/ubuntu16_04/layers/bazel:bazel-tools",
         "//container/ubuntu16_04/layers/bazel:bazel-ltl",
         "//container/ubuntu16_04/layers/git:git-ltl",
+        "//container/ubuntu16_04/layers/wget:wget-ltl",
```
nlopezgi: why not add wget to bazel-tools?

xingao267: I was considering that. The main reason I made it a separate LTL is because there is an Argo wget container (https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/wget). So I made it to live by itself. Same reason for the git LTL.

nlopezgi: sgtm

PR #96, opened 2018-06-20
Group repo dependencies into a macro.

xingao267:
Also fix a bug when autoconfig target is located in the root BUILD file
of the workspace.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -13,28 +13,40 @@
 # limitations under the License.
 workspace(name = "bazel_toolchains")
 
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load(
     "//skylib:package_names.bzl",
     "jessie_package_names",
 )
-
-# Use http_archive rule instead of git_repository rule
-# https://docs.bazel.build/versions/master/be/workspace.html#git_repository
-http_archive(
-    name = "io_bazel_rules_docker",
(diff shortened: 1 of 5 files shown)


rules/docker_config.sh.tpl
```diff
@@ -23,7 +23,7 @@ main() {
   # Expand a tar file with the repo if needed
   %{EXPAND_REPO_CMD}
   # Call the script produced by the docker_build rule to load the image
-  %{LOAD_IMAGE_SH}
+  ./%{LOAD_IMAGE_SH}
```
erain: Do we want to include this change in this PR?

WORKSPACE
```diff
@@ -36,6 +32,22 @@ load(
 
 container_repositories()
 
+load("@io_bazel_rules_go//go:def.bzl", "go_register_toolchains", "go_rules_dependencies")
+
+go_rules_dependencies()
```
erain: It is a pity that we couldn't call this repository rule in our repository rule... Maybe we can update README to indicate how should import bazel-toolchains repo in WORKSPACE as well?

xingao267: Yea. good idea. I forgot to do that. Will do.

rules/docker_config.sh.tpl
xingao267: I'll undo this.

WORKSPACE
xingao267: Updated the instructions in the docker_config.bzl, which is where the instructions located before.

PR #97, opened 2018-06-20
Changed install_pkgs output image tags to more sensical ones

ArthurRab:
Naming restrictions no longer apply because of new updates in
base-images-docker

### DIFF
--- a/test/container/BUILD
+++ b/test/container/BUILD
@@ -90,7 +90,7 @@ install_pkgs(
     name = "docker-prereq-test",
     image_tar = "@debian8//image",
     installables_tar = ":docker-prereq-packages.tar",
-    output_image_name = "docker-prereq-test:intermediate",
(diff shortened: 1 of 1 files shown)


ArthurRab:
Local tests I ran passed. Let me know if there are other tags you want changed, or if you have a better idea for what the tags should be. I just based them off of the rule name.

test/container/BUILD
```diff
@@ -90,7 +90,7 @@ install_pkgs(
     name = "docker-prereq-test",
     image_tar = "@debian8//image",
     installables_tar = ":docker-prereq-packages.tar",
-    output_image_name = "bazel/test/container:docker-prereq-test",
+    output_image_name = "docker-prereq-test:intermediate",
```
xingao267: I think tagging them as latest should be good. They are not really intermediate images.

ArthurRab: Fixed

PR #103, opened 2018-06-22
adding docker sandbox flags to 1.14.1 bazelrc

nlopezgi:
Not needed in the other bazelrc files as these are just to support external docs (https://docs.bazel.build/versions/master/remote-execution-sandbox.html)


xingao267:
1.14.1 --> 0.14.1

xingao267:
It seems you didn't commit it using your git account, you might be using your host machine account so the CLA didn't pass. You might need to reconfig you git setting and create a new PR.

PR #105, opened 2018-06-25
Install clang instead of gcc bazel container.

xingao267:
- Also extract some more common yaml file.


container/ubuntu16_04/layers/bazel/BUILD
```diff
@@ -20,12 +20,12 @@ load("//container/rules:docker_toolchains.bzl", "language_tool_layer")
...
-#  gcc + other tools required by Bazel
+# Tools required by Bazel. Note that we do not add a CC complier here as we
+# will include that in the final container.
 language_tool_layer(
     name = "bazel-tools",
     base = "@ubuntu16_04//image",
     packages = [
-        "g++",
         "patch",
         "software-properties-common",
```
nlopezgi: lets try to just add openssl and any other specific packages that are needed

container/ubuntu16_04/builders/bazel/container.yaml
```diff
@@ -1,27 +1,10 @@
 schemaVersion: "2.0.0"
 
 commandTests:
-- name: 'gcc-version'
-  command: 'gcc'
-  args: ['--version']
-  expectedOutput: ['gcc .*']
 - name: 'check-openssl'
```
nlopezgi: Move this one to java layer

PR #106, opened 2018-06-25
Isolate digests into a single file

l27yao:
- Add "toolchain_containers.bzl" to store digests info of base os image and Python3.
- Replace raw digest info in WORKSPACE file and Python BUILD file.


toolchain_containers.bzl
```diff
@@ -0,0 +1,20 @@
+def toolchain_container_sha256s():
```
erain: Let's avoid top-level skylark code. Could you move this file into `rules` ?

l27yao: Sure.

PR #115, opened 2018-07-05
Update bazel-0.15.0.bazelrc

sadaf-matinkhoo:
Add flags required for remote-cache.

### DIFF
--- a/bazelrc/bazel-0.15.0.bazelrc
+++ b/bazelrc/bazel-0.15.0.bazelrc
@@ -125,31 +125,13 @@ build:docker-sandbox --define=EXECUTOR=remote
 build:docker-sandbox --experimental_docker_verbose
 build:docker-sandbox --experimental_enable_docker_sandbox
 
-# Set up flags to enable local execution using remote cache.
-
-# Enable the remote cache so action results can be shared across machines,
-# developers, and workspaces.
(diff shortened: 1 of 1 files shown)


bazelrc/bazel-0.15.0.bazelrc
```diff
@@ -124,3 +124,33 @@ build:docker-sandbox --genrule_strategy=docker
 build:docker-sandbox --define=EXECUTOR=remote
 build:docker-sandbox --experimental_docker_verbose
 build:docker-sandbox --experimental_enable_docker_sandbox
+
+# Set up flags to enable local execution using remote cache.
+
+# Enable the remote cache so action results can be shared across machines,
+# developers, and workspaces.
+build:remote-cache --remote_cache=remotebuildexecution.googleapis.com
```
nlopezgi: remove the comments for separate blocks and leave all flags as a single block w/o empty lines so that its easy to tell them apart from other configs (most of the comments are not too useful, like the encryption, timeout) others probably need to be documented properly elsewhere (comments related to experimental_strict_action_env and how to use sandboxing in combination with caching probably belong in the remote cache documentation)

sadaf-matinkhoo: Done.

PR #119, opened 2018-07-09
Container release for clang r333878.

xingao267:


container/common/clang.yaml
```diff
@@ -4,7 +4,7 @@ commandTests:
 - name: 'clang-version'
   command: 'clang'
   args: ['--version']
-  expectedOutput: ['clang version 7.0.0.*']
+  expectedOutput: ['clang version 7.0.0 [(]trunk 333878[)]']
```
nlopezgi: any way we can get this value in some variable that is more visible?

xingao267: Done. Now get value from revision.bzl

xingao267:
There is an update in Python published 21hr ago. I'm gonna include that as well.

PR #120, opened 2018-07-09
Moved the toolchain_container rule into a macro

ArthurRab:
This way we can supply an installables tar if we want to.
This makes sure our rebuild and the original build have the same rule
producing them up to the installables_tar (easier to maintain).


container/debian8/builders/rbe-debian8/toolchain_build.bzl
```diff
@@ -0,0 +1,61 @@
...
+            # PATH envvar is a special case, and currently only the one in the
+            # topmost layer is set. So that we override it here to include all.
+            "PATH": "$PATH:/opt/python3.6/bin:/usr/local/go/bin",
+            "LANG": "C.UTF-8",
+            "LANGUAGE": "C.UTF-8",
+            "LC_ALL": "C.UTF-8",
+        },
+        installables_tar = installables_tar,
+        language_layers = [
+            "@bazel_toolchains//container/debian8/builders/rbe-debian8:base-ltl",
```
xingao267: Why adding @bazel_toolchains here?

container/debian8/builders/rbe-debian8/toolchain_build.bzl
```diff
@@ -0,0 +1,61 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load(
+    "//container/rules:docker_toolchains.bzl",
+    "toolchain_container",
+)
+load("@io_bazel_rules_docker//contrib:test.bzl", "container_test")
+
+def toolchain_build(name, installables_tar = None):
```
xingao267: toolchain_build_and_test maybe?

container/debian8/builders/rbe-debian8/toolchain_build.bzl
xingao267: and below for yamls?

container/debian8/builders/rbe-debian8/toolchain_build.bzl
xingao267: Document the usage of this new macro

container/debian8/builders/rbe-debian8/toolchain_build.bzl
ArthurRab: I added it because these things are treated as strings while within the macro, so if it is called from another repo, you need the @bazel_toolchains to make it work. Within the original repo this doesn't actually do anything.

container/debian8/builders/rbe-debian8/toolchain_build.bzl
nlopezgi: This macro is hardcoding all the info for the debian8 image and cannot be reused effectively for anything else. iiuc what you are trying to do, is to enable someone to extend the existing toolchain target, by just adding an installables tar. I think we should allow this, but this approach is not the correct one imo. Lets sync up to try to find a better approach.

ArthurRab:
Discussed offline. No clearly clean solution to our problem. Leaving it as is for now.

PR #127, opened 2018-07-19
Update dependencies.

xingao267:


xingao267:
@nlopezgi Attempted a fix. Not sure if it's the right way or not. but seems we can't use more meaningful index when directly using the implementation function.

xingao267:
Having https://github.com/GoogleContainerTools/base-images-docker/pull/243 out for review

xingao267:
okok... with https://github.com/bazelbuild/rules_docker/commit/a25f00cc14425fda603fce5cb84378b62a3999f6, I need to update base-image-docker with the new py scripts.

Issue #66, opened 2018-06-04
Please move from native to Skylark rules for external git/http repositories


aehlig:
Bazel is planning to move away from native rules for external repositories to
Skylark rules, mainly because bundling every version control system in bazel
does not scale. As part of this effort, we plan to submit
https://bazel-review.googlesource.com/c/bazel/+/55932 soon, which will cause
bazel by default to insist that the skylark versions the git and http rules for
external repositories be used. This can be disabled by a flag for at least the
next 6 months, but we still would appreciate if you could adopt your repository
to use the Skylark versions of those rules soon.

A test run of downstream projects with that flag flip in bazel enabled can
be found at https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/286

For more background, see
- the original announcement https://groups.google.com/d/topic/bazel-discuss/dO2MHQLwJF0/discussion
- and the documentation of the incompatible changes https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#remove-native-git-repository

Thanks.

xingao267:
We should be able to proceed with this as https://github.com/bazelbuild/bazel/issues/5633 is fixed.

mhlopko:
Howdy all,
it looks to me that bazelbuild/bazel#5633 is fixed, and you can fix your project. In the meantime the incompatible flag was flipped in Bazel, and bazel-toolchains are red on our downstream pipeline: https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/531#e1dee82d-4f9f-4c89-98b1-781c0f58de38. I'll disable bazel-toolchains for the meantime, please create issue at https://github.com/bazelbuild/continuous-integration to reenable bazel-toolchains once this issue is fixed. Thank you!

mhlopko:
rules_kotlin are also affected by bazel-toolchains, would you mind updating their WORKSPACE to use a fixed version of bazel-toolchains once this is fixed?

xingao267:
Thanks. Working on fixing this.

xingao267:
Should be fixed. Please reopen if there is any more issue related to this.

I'm also sending a PR to continuous-integration to re-enable ci for bazel-toolchains.

PR #128, opened 2018-07-20
Use skylark version of http_file.

xingao267:
Part of #66.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -13,6 +13,7 @@
 # limitations under the License.
 workspace(name = "bazel_toolchains")
 
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")
 load(
     "//skylib:package_names.bzl",
     "jessie_package_names",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -131,13 +131,13 @@ container_pull(
 http_file(
     name = "debian_docker_gpg",
     sha256 = "1500c1f56fa9e26b9b8f42452a553675796ade0807cdce11975eb98170b3a570",
-    url = "https://download.docker.com/linux/debian/gpg",
+    urls = ["https://download.docker.com/linux/debian/gpg"],
 )
 
 http_file(
     name = "xenial_docker_gpg",
     sha256 = "1500c1f56fa9e26b9b8f42452a553675796ade0807cdce11975eb98170b3a570",
-    url = "https://download.docker.com/linux/ubuntu/gpg",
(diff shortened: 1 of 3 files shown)


xingao267:
Fixed.

xingao267:
Saw similar issue here: https://github.com/bazelbuild/bazel/issues/5633. A lot of other rules enforce name extension check, but the downloaded file of skylark http_file will always be `downloaded`. I will have to create a genrule for each of them to rename. Will send a commit for that soon.

test/configs/debian8_clang_autoconfig_test.sh
```diff
@@ -63,7 +63,8 @@ tar -xf ${TEST_TMPDIR}/debian8-clang-${CONFIG_VERSION}-bazel_${BAZEL_VERSION}-au
 rm -rf ${TEST_TMPDIR}/local_config_cc/tools ${TEST_TMPDIR}/local_config_cc/WORKSPACE
 
 # Unpack the tarball containing published toolchain configs for Bazel 0.10.0 from GitHub.
-tar -xf ${TEST_SRCDIR}/bazel_toolchains_test/file/${COMMIT}.tar.gz -C ${TEST_TMPDIR}
+# The downloaded file of `http_file` rule is hardcoded as `file/downloaded`.
+tar -xf ${TEST_SRCDIR}/bazel_toolchains_test/file/downloaded -C ${TEST_TMPDIR}
```
nlopezgi: I dont think the right solution is to hardcode the downloaded name here, you could instead change the target that this depends on (https://github.com/bazelbuild/bazel-toolchains/blob/master/test/configs/BUILD#L146) from @bazel_toolchains_test//file to third_party/bazel_toolchains_test:rename_file (and create a BUILD file in third_party/bazel_toolchains_test:rename_file that has a genrule to rename).

xingao267:
Ok. So still use native http_file in workspace, except in the repositories.bzl, which is used by bazel_gpg key.

WORKSPACE
```diff
@@ -130,13 +130,13 @@ container_pull(
 http_file(
     name = "debian_docker_gpg",
     sha256 = "1500c1f56fa9e26b9b8f42452a553675796ade0807cdce11975eb98170b3a570",
-    url = "https://download.docker.com/linux/debian/gpg",
+    urls = ["https://download.docker.com/linux/debian/gpg"],
```
nlopezgi: just to confirm, there is no change here, just swapping using url, to urls, but still the native rule, right?

xingao267: No change here. native http_file also have urls.

PR #130, opened 2018-07-23
Improve build.py

ArthurRab:
Added custom bazel-version selection to build.py

Updated yaml configs to support this

Made target maps overidable in build.py

Made build.py usable if called from outside directory as long as it is a subdirectory


container/build.py
```diff
@@ -240,6 +286,8 @@ def parse_arguments():
 
   optional.add_argument(
       "-h", "--help", help="print this help text and exit", action="help")
+  optional.add_argument(
+      "-m", "--map", help="overrides target map file path", type=str, default=None)
```
xingao267: be specific this is a path, and relative or absolute or either

container/build.py
```diff
@@ -99,32 +105,66 @@
...
   target = TYPE_TARGET_MAP[type_]
-  tarball = TYPE_TARBALL_MAP[type_]
+  tarball = None
+  if bucket:
+    tarball = TYPE_TARBALL_MAP[type_]
+
+  # Gets the base directory of the bazel-toolchains repo relative to this 
+  # build.py. This is for referencing yaml files and mounting project to gcloud
+  # THIS NEEDS TO BE UPDATED IF THIS FILE IS MOVED 
+  bazel_toolchains_base_dir = os.path.relpath(os.path.join(__file__, "../.."))
```
xingao267: Add more documentation about this line.

container/build.py
```diff
@@ -16,6 +16,9 @@
 
 """Builds a toolchain container, with Google Cloud Container Builder or locally.
 
+IF THIS SCRIPT IS CALLED FROM OUTSIDE OF THE BAZEL-TOOLCHAINS REPO, THE BAZEL-TOOLCHAINS REPO
+MUST BE A SUBDIRECTORY OF THE OUTER PROJECT
```
nlopezgi: Add example illustrating what the correct setup would be (do you need to checkout the entire project as a subdirectory, how would you then use it?

container/build.py
```diff
@@ -27,38 +30,41 @@
...
   -t TAG, --tag TAG     Docker tag for the image
+  -v BAZEL_VERSION, --bazel_version BAZEL_VERSION
+                        The version of Bazel to build the image with, e.g. 0.15.1
 
 optional arguments:
   -a, --async           Asynchronous execute Cloud Container Builder
   -b BUCKET, --bucket BUCKET
                         GCS bucket to store the tarball of debian packages
   -h, --help            print this help text and exit
+  -m MAP, --map MAP     overrides target map file path
```
nlopezgi: can you explain what the effect of overriding it is (or when it needs to be overriden?)

container/build.py
```diff
@@ -99,32 +105,66 @@
...
+  for dirpath, _, files in os.walk("third_party"):
+    for f in files:
+      full_path = os.path.join(dirpath, f)
+      os.chmod(full_path, 0o644)
+
   if local:
     local_build(type_, package, target)
   else:
-    cloud_build(project_root, project, container, tag, async_, bucket, package, target, tarball)
+    # Gets the yaml relative to this build.py, regardless of what directory it was called from
```
nlopezgi: could we receive this info as an optional argument (or somehow make it easier to maintain it so that the detail that the paths must be kept in sync is more visibly)

container/build.py
```diff
@@ -99,32 +105,66 @@
...
+    # MUST BE UPDATED IF THIS FILE OR THE YAML FILE MOVE
+    config_file = "{}/container/cloudbuild.yaml".format(bazel_toolchains_base_dir)
+    extra_substitution = ",_BUCKET={},_TARBALL={}".format(bucket, tarball)
+    if not bucket:
+      # Gets the yaml relative to this build.py, regardless of what directory it was called from
+      # MUST BE UPDATED IF THIS FILE OR THE YAML FILE MOVE
+      config_file = "{}/container/cloudbuild_no_bucket.yaml".format(bazel_toolchains_base_dir)
```
nlopezgi: same here

container/gcloud_bazel_container/Dockerfile.builder
```diff
@@ -0,0 +1,62 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
nlopezgi: As discussed offline, lets try to remove this dockerfile altogether and install the bazel version we need (+gcloud, if needed) as separate steps in the cloudbuild.yaml

container/build.py
```diff
@@ -99,32 +105,66 @@
...
 
 
-def main(type_, project, container, tag, async_, bucket, local):
+def main(type_, project, container, tag, async_, bucket, local, bazel_version, map = None):
   '''Runs the build. More info in module docstring at the top.
   '''
+  
+  if map: # Override the map values
+    map_module = imp.load_source("map", map)
+    TYPE_PACKAGE_MAP = map_module.TYPE_PACKAGE_MAP
```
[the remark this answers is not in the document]
ArthurRab: Added a couple try/except statements

container/build.py
```diff
@@ -99,32 +105,66 @@
...
+  # THIS NEEDS TO BE UPDATED IF THIS FILE IS MOVED 
+  bazel_toolchains_base_dir = os.path.relpath(os.path.join(__file__, "../.."))
 
   # We need to start the build from the root of the project, so that we can
   # mount the full root directory (to use bazel builder properly).
   os.chdir(project_root)
   # We need to run clean to make sure we don't mount local build outputs
   subprocess.call(["bazel", "clean"])
 
+  # Ensure all BUILD files under /third_party have the right permission.
```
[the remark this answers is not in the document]
ArthurRab: It made sense when I was working on moving things around, but ya no longer seems necessary. Moved it back.

container/build.py
ArthurRab: Mentioned it can be either.

container/build.py
ArthurRab: Added a better explanation

container/build.py
ArthurRab: Changed it to find the base of bazel-toolchains using the "git rev-parse --show-toplevel" command, so the location of the build.py no longer matters.

container/build.py
ArthurRab: Fix mentioned above

container/build.py
```diff
@@ -115,32 +119,54 @@
...
+  #     subdir/
+  #         bazel-toolchains
+  #     ...
+  # 
+  # Then if we call build.py with the terminal running in foo,
+  # bazel_toolchains_base_dir == "subdir/bazel-toolchains"
+  # (This also allows for renaming of the bazel-toolchains folder as the variable
+  # will store the path regardless of the directory names) 
+  os.chdir(os.path.dirname(__file__))
+  bazel_toolchains_base_dir = os.path.relpath(subprocess.check_output(
```
xingao267: just to confirm the git root obtained here is the git_root of bazel-toolchains repo right?

container/build.py
xingao267: you should be clear that it's taking a path, and will override TYPE_PACKAGE_MAP, TYPE_TARGET_MAP, TYPE_TARBALL_MAP

container/build.py
```diff
@@ -189,10 +215,21 @@ def cloud_build(project, container, tag, async_, package, target,
       extra_substitutions: any extra substitutions required for the given yaml file
                             mainly used for _BUCKET and _TARBALL
   '''
+
+  # We need to run clean to make sure we don't mount local build outputs
+  subprocess.call(["bazel", "clean"])
```
xingao267: should clean before both local and cloud build?

container/build.py
```diff
@@ -99,32 +108,89 @@
...
   # mount the full root directory (to use bazel builder properly).
   os.chdir(project_root)
-  # We need to run clean to make sure we don't mount local build outputs
-  subprocess.call(["bazel", "clean"])
+
+  # Ensure all BUILD files under /third_party have the right permission.
+  # This is because in some systems the BUILD files under /third_party
+  # (after git clone) will be with permission 640 and the build will
+  # fail in Container Builder.
+  for dirpath, _, files in os.walk("third_party"):
```
xingao267: This should stay in cloud_build function.

container/build.py
```diff
@@ -99,32 +108,89 @@
...
-    cloud_build(project_root, project, container, tag, async_, bucket, package, target, tarball)
+    # Gets the yaml relative to the bazel-toolchains root, regardless of what directory it was called from
+    # MUST BE UPDATED IF THE YAML FILE IS MOVED
+    config_file = "{}/container/cloudbuild.yaml".format(bazel_toolchains_base_dir)
+    extra_substitution = ",_BUCKET={},_TARBALL={}".format(bucket, tarball)
+    if not bucket:
+      # Gets the yaml relative to the bazel-toolchains root, regardless of what directory it was called from
+      # MUST BE UPDATED IF THE YAML FILE IS MOVED
+      config_file = "{}/container/cloudbuild_no_bucket.yaml".format(bazel_toolchains_base_dir)
+      extra_substitution = ""
```
xingao267: This part should also stay in cloud_build function. Only move things around if necessary.

container/build.py
ArthurRab: That's where it was originally called before, but it does only seem necessary for the cloud one. Moved

container/build.py
ArthurRab: Moved it back

container/build.py
ArthurRab: Moved it back

container/build.py
```diff
@@ -221,6 +334,13 @@ def parse_arguments():
   required.add_argument(
       "-t", "--tag", help="Docker tag for the image", type=str)
 
+  required.add_argument(
+      "-v",
+      "--bazel_version",
+      help="The version of Bazel to build the image with, e.g. 0.15.1",
```
xingao267: ...on Google Cloud Container Builder.

container/build.py
```diff
@@ -26,45 +42,49 @@
...
-  -d TYPE, --type TYPE  Type of the container: see TYPE_TARGET_MAP
+  -d, --type            Type of the container: see SUPPORTED_TYPES
   -p PROJECT, --project PROJECT
                         GCP project ID
   -c CONTAINER, --container CONTAINER
                         Docker container name
   -t TAG, --tag TAG     Docker tag for the image
+  -v BAZEL_VERSION, --bazel_version BAZEL_VERSION
+                        The version of Bazel to build the image with, e.g.
+                        0.15.1
```
xingao267: .... on Google Cloud Container Builder

container/build.py
```diff
@@ -26,45 +42,49 @@
 $ python build.py -d {container_type} -l
 will produce docker locally as {container_type}:latest
 
-usage:
-  build.py -d {rbe-debian8,rbe-debian9,rbe-ubuntu16_04,ubuntu16_04-bazel-docker,ubuntu16_04-bazel}
-      [-p PROJECT] [-c CONTAINER] [-t TAG] [-a] [-b BUCKET] [-h]
-      [-l]
-
 required arguments:
```
xingao267: This is not very accurate given some arguments are not required for local build.

container/build.py
```diff
@@ -26,45 +42,52 @@
...
-  -d TYPE, --type TYPE  Type of the container: see TYPE_TARGET_MAP
+  -d, --type            Type of the container: see SUPPORTED_TYPES
+required arguments (for cloud build):
   -p PROJECT, --project PROJECT
                         GCP project ID
   -c CONTAINER, --container CONTAINER
                         Docker container name
   -t TAG, --tag TAG     Docker tag for the image
+  -v BAZEL_VERSION, --bazel_version BAZEL_VERSION
```
nlopezgi: does this also work with rcX versions of bazel (if so add example, if not make explicit)?

container/build.py
```diff
@@ -26,45 +42,52 @@
...
 optional arguments:
+  -h, --help            print this help text and exit
+
+optional arguments (for cloud build):
   -a, --async           Asynchronous execute Cloud Container Builder
   -b BUCKET, --bucket BUCKET
                         GCS bucket to store the tarball of debian packages
-  -h, --help            print this help text and exit
+  -m MAP, --map MAP     path to override target map file (can be absolute or
+                        relative). This allows the building of bazel targets
```
nlopezgi: r/building

container/build.py
```diff
@@ -13,11 +13,27 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""usage: build.py -d TYPE [-p PROJECT] [-c CONTAINER] [-t TAG]
+                   -v BAZEL_VERSION [-a] [-b BUCKET] [-h] [-m MAP] [-l]
 
-"""Builds a toolchain container, with Google Cloud Container Builder or locally.
+Builds the fully-loaded container, with Google Cloud Container Builder or
```
nlopezgi: Builds a toolchain container

container/build.py
```diff
@@ -13,11 +13,27 @@
...
+locally.
+
+IF THIS SCRIPT IS CALLED FROM OUTSIDE OF THE BAZEL-TOOLCHAINS REPO, THE
```
nlopezgi: The issue I have with these docs is I would not know why I would call this script from a different repo. But this might be too much to include here. We might need a separate README to explain that better.

container/build.py
```diff
@@ -26,45 +42,52 @@
...
-                    "rbe-ubuntu16_04",
-                    "ubuntu16_04-bazel",
-                    "ubuntu16_04-bazel-docker"
-                  ]
+SUPPORTED_TYPES = [
+    "rbe-debian8", "rbe-debian9", "rbe-ubuntu16_04", "ubuntu16_04-bazel",
+    "ubuntu16_04-bazel-docker"
+]
+
+# ALL FILES USED IN -m TO OVERRIDE TARGETS MUST BE IN THIS FORM
```
nlopezgi: File passed in -m must include the following 3 maps (all 3 with the same keys, and corresponding values):

container/build.py
nlopezgi: path (can be absolute or relative) to file containing 3 maps to
override the default ones defined below (TYPE_PACKAGE_MAP, TYPE_TARGET_MAP, and TYPE_TARBALL_MAP)

container/build.py
ArthurRab: Added

container/build.py
ArthurRab: Changed it to use this help text

container/build.py
ArthurRab: It only supports versions supported by our bazel rules, added a reference to the version file.

container/build.py
ArthurRab: Fixed

container/build.py
ArthurRab: Discussed offline. README should come in a future pr

PR #133, opened 2018-07-25
Update latest Bazel version.

xingao267:
Also use subprocess.check_call instead of subprocess.call. The former
will exit as soon as the subcommand exits with non-zero status.


xingao267:
@ArthurRab FYI, i'm changing to subprocess.call to subprocess.check_call in this PR. You might need to fix merging conflicts when you rebase on it.

PR #138, opened 2018-07-26
Stop using DefaultInfo.runfiles

c-parsons:
Use DefaultInfo.default_runfiles instead of DefaultInfo.runfiles, as DefaultInfo.runfiles is an undocumented API that is being completely eliminated in the current Bazel release (0.16.0).


xingao267:
Thanks for sending the fix! Could you also change https://github.com/bazelbuild/bazel-toolchains/blob/master/rules/docker_config.bzl#L283?

c-parsons:
Done!

PR #142, opened 2018-07-30
Add gcloud layer.

xingao267:


container/ubuntu16_04/layers/gcloud/gcloud.yaml
```diff
@@ -0,0 +1,7 @@
+schemaVersion: "2.0.0"
+
+commandTests:
+- name: 'check-gcloud'
+  command: 'gcloud'
+  args: ['version']
+  expectedOutput: ['Google Cloud SDK.*']
```
nlopezgi: Can you create a bzl file with the gcloud sha256 + version string and use a tpl to generate a yaml file that checks the version (and replace use of sha in WORKSPACE for use of bzl)

xingao267: done

The sha in WORKSPACE is for the pgp key, which doesn't change normally.

PR #153, opened 2018-08-07
Add test config to run against RBE.

xingao267:


xingao267:
docker command is not yet supported.

PR #163, opened 2018-08-13
Add new WORKSPACE rule to download gcs objects

l27yao:


rules/gcs.bzl
```diff
@@ -0,0 +1,91 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Skylark rule for pulling a file from GCS bucket.
+"""
+
+_GCS_FILE_BUILD = """
+package(default_visibility = ["//visibility:public"])
```
erain: I think this line should be the first line after the comments

rules/gcs.bzl
```diff
@@ -0,0 +1,91 @@
...
+        ctx.path("file/BUILD"),
+        ctx.path("file/BUILD.bazel"),
+    ]
+    downloaded_file_path = ctx.attr.downloaded_file_path or ctx.attr.file
+    download_path = ctx.path("file/" + downloaded_file_path)
+    if download_path in forbidden_files or not str(download_path).startswith(str(repo_root)):
+        fail("'%s' cannot be used as downloaded_file_path in gcs_file" % ctx.attr.downloaded_file_path)
+
+    # Add a top-level BUILD file to export all the downloaded files.
+    ctx.file("file/BUILD", _GCS_FILE_BUILD.format(downloaded_file_path))
```
erain: AFAIK quite some different rules would use the folder `file`. Since we are dealing with GCS file here so let's be more specific and call it `gcsfile` and the build file `gcsfile/BUILD`

rules/gcs.bzl
```diff
@@ -0,0 +1,91 @@
...
+        "{bucket}/{file}".format(
+            bucket = ctx.attr.bucket,
+            file = ctx.attr.file,
+        ),
+        download_path,
+    ]
+    gsutil_cp_result = ctx.execute(gsutil_cp_cmd)
+    if gsutil_cp_result.return_code:
+        fail("gsutil cp command failed: %s (%s)" % (gsutil_cp_result.stderr, " ".join(gsutil_cp_cmd)))
+    ctx.file("validation.sh", """
```
erain: Instead of have `validation.sh` file inside of the skylark logic, could we make it a standalone shell script and we can execute it by `bash validation.sh FILENAME SHA256`?

rules/gcs.bzl
```diff
@@ -0,0 +1,91 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Skylark rule for pulling a file from GCS bucket.
+"""
+
+_GCS_FILE_BUILD = """
```
erain: This looks a little bit odd to me... could you give me some reference why we need to do `_GCS_FILE_BUILD=...`?

rules/gcs.bzl
nlopezgi: this is the content of the produced BUILD file, so needs to be inside this variable. The alternative (which might be good) is to have this in a template file, and use repository_ctx.template (https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#template)

rules/gcs.bzl
nlopezgi: this should be fine as file. This repo is created by this rule, and no other rules should create any files under this folder.

rules/gcs.bzl
nlopezgi: lets make a template file that both executes the gsutil cp command and validates its sha, and then, create the file via repository_ctx.template (https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#template). That way you need to call only ctx.execute once (you may not be able to remove the file afterwards, but I don't think that's a problem)

l27yao: Made a template file for creating bash script in the new commit.

rules/gcs.bzl
l27yao: I will keep content of BUILD file here, but can move it to a template if we want.

rules/gcs.bzl
l27yao: I referred to this: https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/http.bzl#L58

PR #167, opened 2018-08-16
adding GCOV env var to clang layers

nlopezgi:
* Use of var will remove warning in bazel builds
* also refactor env vars for clang to a separate package


nlopezgi:
abandoning in favor of https://github.com/bazelbuild/bazel/pull/5946

xingao267:
This needs to be revived. I'm creating a PR for this. See https://github.com/bazelbuild/bazel/commit/1d956c707e1c843896ac58a341c335c9c149073d.

I have experienced this when I try to use Bazel 0.17.1 in a env without GCOV set

PR #170, opened 2018-08-17
Add experimental trusty based Bazel container.

xingao267:


WORKSPACE
```diff
@@ -127,3 +127,17 @@ http_file(
         "https://github.com/bazelbuild/bazel-toolchains/archive/44200e0c026d86c53470d107b3697a3e46469c43.tar.gz",
     ],
 )
+
+container_pull(
```
nlopezgi: Move this to images.bzl
Add note we pull trusty base from ubuntu repo (and not from l.gcr.io as we do for other base containers)

xingao267: Done.

PR #172, opened 2018-08-21
Use pre-built bazel-docker-gcloud container.

xingao267:


container/cloudbuild.yaml
```diff
@@ -19,27 +19,17 @@
...
-    - '--output_base=/workspace'
-    - 'run'
-    - '${_BAZEL_TOOLCHAINS_REF}//container/ubuntu16_04/builders/bazel_docker_gcloud:bazel_${_BAZEL_VERSION}_docker_gcloud'
-    id: 'builder'
-    waitFor:
-    - "-" # wait for nothing - start immediately
-
 # Step: print the bazel version for confirmation
-  - name: 'bazel/container/ubuntu16_04/builders/bazel_docker_gcloud:bazel_${_BAZEL_VERSION}_docker_gcloud'
+  - name: 'gcr.io/asci-toolchain/nosla-ubuntu16_04-bazel-docker-gcloud:${_BAZEL_VERSION}'
```
nlopezgi: this would mean that effectively the argo build can only be executed from the asci-toolchains projects to have access to the private container, right?

xingao267: Yes

nlopezgi: Ok, lets change https://github.com/bazelbuild/bazel-toolchains/blob/master/container/README.md to make explicit this process can currently only be executed if by users who have access to asci-toolchains gcr but we publish it here to make the build process for our containers publicly available. Also add pointers here to how we build each of these containers (e.g., just pointing to the target in this repo we used to build the container should be sufficient?)

xingao267: Updated. Didn't list individual targets, but pointing users to look at build locally section instead.

container/README.md
```diff
@@ -25,6 +25,11 @@ Bazel](https://github.com/bazelbuild/bazel/tree/master/src/main/java/com/google/
 
 ### Build with [Google Cloud Container Builder](https://cloud.google.com/container-builder/)
 
+**Note: currently, this process can currently only be executed by users who have
```
nlopezgi: r/ second currently

container/README.md
```diff
@@ -25,6 +25,11 @@ Bazel](https://github.com/bazelbuild/bazel/tree/master/src/main/java/com/google/
...
+read access to asci-toolchains's GCR, but we publish it here to make the build
```
nlopezgi: GCR. We publish...

PR #173, opened 2018-08-23
Do not make wget and git as separate layers.

xingao267:


container/debian8/builders/bazel/BUILD
```diff
@@ -20,6 +20,15 @@ load("//container/rules:docker_toolchains.bzl", "toolchain_container")
 load("@io_bazel_rules_docker//contrib:test.bzl", "container_test")
 load("//container/common/bazel:version.bzl", "BAZEL_VERSION_SHA256S")
 
+language_tool_layer(
```
nlopezgi: maybe we should put these in bazel-tools layer?

xingao267: I thought about that. But I feel these tools are not needed for installing or running Bazel, but help tools we need to the container.

nlopezgi: lets leave it as per conversation offline. Lets call it extra-tools-ltl and have a comment explaining this is a bucket for all tools that are needed in most builds, but not strictly required by bazel

PR #177, opened 2018-09-05
Fix builtin include directory for clang 8.

iphydf:


nlopezgi:
Thanks for your contribution.
Our current container that supports bazel 0.16.1 still has clang 7.0
Our new container (coming out soon) will come with clang 8.0. We will update the config files accordingly once the container is available to the public in http://gcr.io/cloud-marketplace/google/rbe-ubuntu16-04

xingao267:
Hi iphydf@, thanks for sending the PR. These configs are auto-generated and we do not take external contributions for configs under configs/ folder. 

We have noticed that there is a clang major version update from 7 to 8. We are working on publishing the new container and configs for that. Thanks again.

PR #179, opened 2018-09-05
Update clang_debian8 and clang_ubuntu16_04 base images (r340178)

xingao267:
Also pin the container used by
//test/configs:debian8_clang_autoconfig_test to a specific digest.


xingao267:
I have another option which is to generate the new configs in the same PR, and update the test to use the new configs.

But in 6 month, this would break again when clang is updated to 0.9.0.

test/configs/BUILD
```diff
@@ -136,13 +137,26 @@ docker_toolchain_autoconfig(
     test = True,
 )
 
+# Container: gcr.io/cloud-marketplace/google/clang-debian8@sha256:ac3b1fdc22c0f2b95abe67f2daf33788425fab52d4e6845900bfe1a42443098f
+docker_toolchain_autoconfig(
```
nlopezgi: What's the purpose of this test? To make sure we can still run autoconfig with the base container adding all repos? If so, please add comment.

xingao267: This is to make sure the autoconfig rule in the current state of the repo can still reproduce the same configs as those published before. We should probably bump the config and bazel version used in the test. But I wanna do that after this release and use the latest one.

PR #180, opened 2018-09-05
Config release for new Clang r340178 containers (partial).

xingao267:
This commit only includes the new cc and java configs for the new
container. The sample toolchain.bazelrc, sample bazelrc file and
platform targets will be added once the new containers are publically
available.


configs/ubuntu16_04_clang/2.0/BUILD
```diff
@@ -0,0 +1,32 @@
+# Copyright 2016 The Bazel Authors. All rights reserved.
```
nlopezgi: Lets create a 1.1 ubuntu16_04_clang version not a 2.0

xingao267: Done.

PR #183, opened 2018-09-13
Set entrypoint to bazel in bazel containers.

xingao267:


container/debian8/builders/bazel/BUILD
```diff
@@ -59,6 +55,18 @@ language_tool_layer(
...
 
+# Set up `entrypoint` and cmd in the container. These cannot be down in the
+# `toolchain_container` rule as they would be consumed by intermediate
+# `container_image` rule in its implementation.
+# TODO(xingao): enable setting things like `entrypoint` and `cmd` in
+# `toolchain_container`.
+[container_image(
+    name = "bazel_%s" % bazel_version,
+    base = ":bazel_%s_intermediate" % bazel_version,
+    entrypoint = ["bazel"],
```
nlopezgi: we should probably use absolute path to binary (lang_image rules all use full path)

PR #187, opened 2018-09-14
Add a check for GCOV env variable.

xingao267:


container/ubuntu14_04/layers/gcc/gcc.yaml
```diff
@@ -5,3 +5,8 @@ commandTests:
   command: 'g++'
   args: ['--version']
   expectedOutput: ['g[+][+] [(]Ubuntu .*']
+
+metadataTest:
+  env:
+    - key: 'GCOV'
```
nlopezgi: I think gcov can be set to something useful with GCC, but not with clang, you can probably check in the container just by trying to find files with name *gcov*

PR #195, opened 2018-10-03
refactoring rule code

nlopezgi:
* moving all rule code under /rules
* use skylib_library to define deps between bzl files
* add explicit references to @bazel_toolchains in load statements in rules that can be reused from outside this repo


rules/BUILD
```diff
@@ -31,3 +32,35 @@ pkg_tar(
         "cc-sample-project/test.cc",
     ],
 )
+
+skylark_library(
```
xingao267: Does this affect how environments.bzl is used? Before we were calling 
load("@bazel_toolchains//rules:environments.bzl", "clang_env")

is it now

load("@bazel_toolchains//rules:environment", "clang_env")?

rules/docker_config.bzl
```diff
@@ -118,7 +118,7 @@ load(
     "@io_bazel_rules_docker//container:container.bzl",
```
xingao267: Move this file to rules/config/docker_config.bzl?

xingao267: And we need to help update abseil's project for this.

test/configs/BUILD
```diff
@@ -16,7 +16,7 @@ licenses(["notice"])  # Apache 2.0
 
 package(default_visibility = ["//visibility:public"])
 
-load("//container/rules:docker_toolchains.bzl", "toolchain_container")
+load("//rules/container:docker_toolchains.bzl", "toolchain_container")
```
xingao267: Add @bazel_toolchains in all these as well?

test/container/BUILD
```diff
@@ -19,7 +19,7 @@ package(default_visibility = ["//visibility:public"])
 load("@base_images_docker//package_managers:download_pkgs.bzl", "download_pkgs")
 load("@base_images_docker//package_managers:install_pkgs.bzl", "install_pkgs")
 load("@base_images_docker//package_managers:apt_key.bzl", "add_apt_key")
-load("//container/rules:docker_toolchains.bzl", "toolchain_container")
+load("//rules/container:docker_toolchains.bzl", "toolchain_container")
```
xingao267: @bazel_toolchains

repositories/BUILD
```diff
@@ -15,3 +15,22 @@
 licenses(["notice"])  # Apache 2.0
 
 package(default_visibility = ["//visibility:public"])
+
+load("@bazel_skylib//:skylark_library.bzl", "skylark_library")
```
xingao267: https://github.com/bazelbuild/bazel-skylib/blob/master/skylark_library.bzl#L1

rules/BUILD
nlopezgi: No it does not affect how load statements are used, I think it affects mainly how the files are used to generate all the documentation for a set of skylark rules (https://skydoc.bazel.build/docs/getting_started.html) - which we should also do soon, and for tests - https://github.com/bazelbuild/bazel-skylib#bzl_library (though I honestly don't know much about tests that require use of this, but will find out more if I can).

repositories/BUILD
nlopezgi: yes, its being deprecated, but the latest release (0.5.0) does not include this change yet, so will do that in upcoming PR once the latest release includes it.

rules/docker_config.bzl
nlopezgi: I dont want to move docker_config precisely for that reason, lets talk about that one more to see if we do it now (though not in this PR), later, or leave it as it is now.

rules/BUILD
xingao267: Ack.

repositories/BUILD
xingao267: Ack.

rules/docker_config.bzl
xingao267: I'm ok with leaving it under rules/ as it is. We probably will create a new repository rule version and that can go under rules/config/.

test/configs/BUILD
nlopezgi: done

test/container/BUILD
nlopezgi: done

WORKSPACE
```diff
@@ -14,7 +14,7 @@
 workspace(name = "bazel_toolchains")
 
 load(
-    "//skylib:package_names.bzl",
+    "//rules/skylib:package_names.bzl",
```
xingao267: @bazel_toolchains

nlopezgi: WORKSPACE files cannot be referenced from another repo. We're adding @bazel_toolchains only to targets that we expect will be used from outside this repo (mainly the docker autoconfig rule, container creation rules and container layers/builders) so that it's possible to do so w/o issues.

xingao267: Sorry... didn't realize it's the WORKSPACE file.

PR #199, opened 2018-10-12
Update rules_docker pin to new commit

smukherj1:


smukherj1:
Don't review just yet. I want to see what breaks

Issue #198, opened 2018-10-12
Bazel@HEAD broke toolchains


mhlopko:
Hi all,

tonight downstream pipeline was red because of toolchain_identifier error:  https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/498#d982b428-5036-41e2-9edc-1107b286d850

This might have been caused by https://github.com/bazelbuild/bazel/commit/c3fb1db9e4e817e8a911f5b347b30f2674a82f7c or by (now rolledback) https://github.com/bazelbuild/bazel/commit/3aedb2f6de80630f88ffb6b60795c44e351a5810 (leaving for the plane in 3 minutes, cannot investigate more).

Can you help me understand what caused the failure, and whether you think the new behavior should be guarded with an incompatible flag (https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#disallow-using-crosstool-to-select-the-cc_toolchain-label).

xingao267:
It's probably due to bazelbuild/bazel@c3fb1db. The error message is "ERROR: Toolchain identifier 'local' was not found, valid identifiers are [stub_armeabi-v7a, linux_gnu_x86, msys_x64_mingw, msvc_x64]".

The tests that are failing are doing a simple cc autoconfig using Bazel@HEAD in a container (doing `bazel build --all_incompatible_changes @local_config_cc//...` using bazel@HEAD in a container).

I'm trying to understand why "local" is no longer a valid toolchain identifier.

mhlopko:
Cc @meteorcloudy. Ha I finally know what's going on. I'm rolling back the bazelbuild/bazel@c3fb1db, but the fix has to be made in bazel-toolchains. The problem is that you're building @local_config_cc//..., but with a different CROSSTOOL (because you specify your own toolchain using --crosstool_top). That results in @local_config_cc//tools/cpp:cc-compiler-k8 to be used with a different CROSSTOOL, and failing. I consider building cc_toolchain with a different CROSSTOOL to be unsupported use case and therefore outside of backwards incompatible policy. And FYI this would just work as expected if local_config_cc used (cc_toolchain_suite|cc_toolchain).proto attribute, or when we'll migrate to crosstool in skylark (https://github.com/bazelbuild/bazel/issues/5380).

Today it works just because toolchain selection is super benevolent and will accept anything that even remotely looks like could be used. I'm fixing that. I'm sure that's not what you intended to test.

[...]

PR #203, opened 2018-10-16
temporarily fix these tests by using incompatible_changes_off

nlopezgi:
https://github.com/bazelbuild/bazel-toolchains/issues/198


xingao267:
Close if favor of #205

PR #208, opened 2018-10-17
Fixing issues with latest changes to docker autoconfig

nlopezgi:
Fixing release script
fixing rule when multiple external repos are pulled from the container


release/cc/execute_targets.py
```diff
@@ -83,16 +83,22 @@ def execute_and_extract_configs(container_configs_list, bazel_version):
...
 
-      command = ("bazel run --define=DOCKER_AUTOCONF_OUTPUT={OUTPUT_DIR} "
-                 "//{PACKAGE}:{TARGET}").format(
+      command = ("bazel build //{PACKAGE}:{TARGET}").format(
+                     PACKAGE=container_configs.package,
+                     TARGET=target)
+      print("\nExecuting command: %s\n" % command)
+      subprocess.check_call(shlex.split(command))
+
+      command = ("cp ./bazel-out/k8-fastbuild/bin/{PACKAGE}/{TARGET}_outputs.tar "
```
xingao267: Will this make the tarball left in the {OUTPUT_DIR}, if so, we should delete it after extraction.

tests/config/BUILD
```diff
@@ -108,6 +108,19 @@ docker_toolchain_autoconfig(
     test = True,
 )
 
+docker_toolchain_autoconfig(
```
xingao267: Can we add the test for this to presubmit?

xingao267:
Let's also include this PR in the coming release (release for bazel 0.18.0) as it fixes autoconfig.

tests/config/BUILD
nlopezgi: done

nlopezgi:
added new test.
+1 to waiting for this PR to push the new release

release/cc/execute_targets.py
nlopezgi: the existing functionality was to output the tar to the OUTPUT_DIR
I'm just keeping it on par by copying it to OUTPUT_DIR after bazel build
 (i.e., this should already be cleaned up by the existing clean up process)

PR #210, opened 2018-10-18
Fix add_apt_repository shebang in containers.

xingao267:


xingao267:
ummmmm, I just realized that software-properties-common is not installed in our containers, and add-apt-repository is part of that. We cannot edit that file before shipping our containers. Or we could install software-properties-common in our containers.

nlopezgi:
lets install it

nlopezgi:
hmm , software properties common includes lots of things. Can we narrow down what package is needed for this specific issue.

xingao267:
Looks like a bunch of locale files https://packages.debian.org/sid/all/software-properties-common/filelist. The total installed size is not big though, 458.0 kB

xingao267:
But it does depend on a lot of things. Let me see.

PR #220, opened 2018-10-25
change python3 version in bazel containers

nlopezgi:
* The py3 version in rbe containers is problematic for add-apt-repo commands
* For the bazel containers that dont require as an up to date version of py3 use apt-get to install that.

### DIFF
--- a/container/common/python3.yaml
+++ b/container/common/python3.yaml
@@ -18,5 +18,5 @@ commandTests:
   expectedOutput: ['Python 3.*.*']
 - name: 'apt-add-repo-check'
   command: 'apt-add-repository'
-  args: ['multiverse']
-  expectedOutput: ['.* distribution component is already enabled for all sources.']
+  args: ['-h']
+  expectedOutput: ['Usage: apt-add-repository']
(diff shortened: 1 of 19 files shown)


container/common/python3.yaml
```diff
@@ -15,4 +15,4 @@ commandTests:
 - name: 'python3-version'
```
xingao267: We can add a new command test to test add_apt_repository

container/debian8/builders/rbe-debian8/BUILD
```diff
@@ -65,7 +65,7 @@ toolchain_container(
         "//container/debian8/layers/go:go-ltl",
         "//container/debian8/layers/java:java-ltl",
         "//container/debian8/layers/java:java10-ltl",
-        "//container/debian8/layers/python:python-ltl",
+        "//container/debian8/layers/python:python-rbe-ltl",
```
xingao267: You probably also need to update the internal container reproduction test after this change.

nlopezgi: thanks for the reminder

container/common/python3.yaml
nlopezgi: done

PR #223, opened 2018-10-26
Produce docker autoconfig logs (#1)

nlopezgi:
Docker autoconfig is not printing out anywhere the results of executing commands inside the container. This PR adds a new output to the rule that produces the log.
Example outputs produced by an autoconfig rule:
Target //tests/config:ubuntu-xenial-autoconfig up-to-date:
bazel-bin/tests/config/ubuntu-xenial-autoconfig-layer.tar
bazel-bin/tests/config/ubuntu-xenial-autoconfig.tar
bazel-bin/tests/config/ubuntu-xenial-autoconfig.digest
bazel-bin/tests/config/ubuntu-xenial-autoconfig_log.log
bazel-bin/tests/config/ubuntu-xenial-autoconfig_outputs.tar

Also added print message in rule that shows instructions for reproducing
execution of docker container. Rule prints out (example):

== Docker autoconfig will run. ==
To debug any errors run:
docker run -d <image_id> bash
Where <image_id> is the image id printed out by the ubuntu-xenial-autoconfig_extract.tar rule.
Then run:
/ubuntu-xenial-autoconfig_install.sh
from inside the container.

### DIFF
--- a/rules/docker_config.bzl
+++ b/rules/docker_config.bzl
@@ -152,6 +152,7 @@ def _docker_toolchain_autoconfig_impl(ctx):
     bazel_config_dir = "/bazel-config"
     project_repo_dir = "project_src"
     name = ctx.attr.name
+    outputs_tar = ctx.outputs.output_tar.basename
 
     # Command to retrieve the project from github if requested.
     clone_repo_cmd = "cd ."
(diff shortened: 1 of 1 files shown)


rules/docker_config.bzl
```diff
@@ -295,6 +310,7 @@ docker_toolchain_autoconfig_ = rule(
         ),
     },
     outputs = _container.image.outputs + {
+        "log": "%{name}_log.log",
```
xingao267: why not just %{name}.log

rules/docker_config.bzl
```diff
@@ -256,14 +256,29 @@ def _docker_toolchain_autoconfig_impl(ctx):
...
         workdir = bazel_config_dir,
     )
-
+    
+    # Commands to run script to create autoconf results, output stderr to log file
+    # add the log file to a tar file and append the output.tar to that same tar file
+    commands = []
+    commands+= ["/" + ctx.attr.name + "_install.sh 2> /" + ctx.attr.name + "_log.log"]
+    commands+= ["tar -cf /extract.tar /" + ctx.attr.name + "_log.log"]
+    commands+= ["tar -rf /extract.tar /" + ctx.attr.name + "_outputs.tar"]
```
xingao267: Combine line 264 and 265 in one command?

nlopezgi: I split them to be able to deal with errors in case outputs.tar was not produced. Updated code to include proper handling of error case (i.e., no failure is produced in extract rule if outputs.tar is not produced)

rules/docker_config.bzl
nlopezgi: done

PR #235, opened 2018-11-08
Add new constraint and toolchain for docker support.

xingao267:
https://github.com/bazelbuild/continuous-integration/issues/376


configs/debian8_clang/latest/BUILD
```diff
@@ -52,3 +52,21 @@ alias(
     name = "javabase",
     actual = PACKAGE + LATEST_CONFIG_VERSION + ":jdk8",
 )
+
+# For internal testing purpose only.
+toolchain(
+    name = "toolchain_docker",
+    exec_compatible_with = [
+        "@bazel_tools//platforms:linux",
```
nlopezgi: can you verify if we _really_ need the exec_compatible_with given comments in https://github.com/bazelbuild/rules_k8s/issues/215

xingao267: In this case I think we do need the exec_compatible_with constraints in order to select the right execution platform to run remote actions.

xingao267: Especially the new support_docker constraint is used to select the container with docker installed.

PR #237, opened 2018-11-12
update bazel version to test in autoconfig

nlopezgi:


xingao267:
This is failing because without change https://github.com/bazelbuild/bazel/commit/3c7c273beb71f0533abd399d574476fa6bee5d02, by changing the order of the flags in https://github.com/bazelbuild/continuous-integration/pull/378, flags that come after will override flags that come before. Some the wrong platform is selected.

meteorcloudy:
@xingao267 Can I assume `--extra_toolchains, --extra_execution_platforms, --extra_execution_platforms, and --platforms` will always appear together in the .yml file? Or should I override them individually?

xingao267:
@meteorcloudy I think currently in remote execution case, we always use --extra_execution_platforms, --host_platform and --platforms together and assign the same value to them. But --extra_toolchains should be considered individually.

meteorcloudy:
Does it make sense to remove them from bazelci.py and make sure users specify them in .yml file explicitly?

xingao267:
Umm, I think it would be too much an overhead for repo owners to have to add them to their yaml files. In 90% cases, people are using the default platform and toolchains.

meteorcloudy:
Understand, then I'll try to implement the override logic in bazelci.py

Issue #239, opened 2018-11-16
Bazel toolchain is failing on CI with Bazel release


meteorcloudy:
https://buildkite.com/bazel/bazel-toolchains/builds/979
```

ERROR: /var/lib/buildkite-agent/builds/bk-worker-ubuntu1604-java8-rtdd/bazel/bazel-toolchains/tests/config/BUILD:43:1: Couldn't build file tests/config/debian-jessie-autoconfig_extract.tar: SkylarkAction tests/config/debian-jessie-autoconfig_extract.tar failed (Exit 1): debian-jessie-autoconfig_extract.build failed: error executing command
--
  | (cd /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/3a0e82dddcec7ab2f14882b3c8a552de/execroot/bazel_toolchains && \
  | exec env - \
  | bazel-out/k8-fastbuild/bin/tests/config/debian-jessie-autoconfig_extract.build)
  | ++ python external/io_bazel_rules_docker/contrib/extract_image_id.py bazel-out/k8-fastbuild/bin/tests/config/debian-jessie-autoconfig.tar
  | + image_id=866092309ed7fed003e7ddd4c4f54f35ab21a08aac60967baa2519638a9a6dc5
  | + docker load -i bazel-out/k8-fastbuild/bin/tests/config/debian-jessie-autoconfig.tar
  | Loaded image: bazel/tests/config:debian-jessie-autoconfig
  | ++ docker run -d 866092309ed7fed003e7ddd4c4f54f35ab21a08aac60967baa2519638a9a6dc5 sh -c '/debian-jessie-autoconfig_install.sh 2> /debian-jessie-autoconfig.log && tar -cf /extract.tar /debian-jessie-autoconfig.log && if [ -f /debian-jessie-autoconfig_outputs.tar ]; then tar -rf /extract.tar /debian-jessie-autoconfig_outputs.tar; fi'
  | + id=03896b2bb2375fcd45efe0f1b64e95d417846dd804da9fe93d76ecf6f19c74b4
  | + docker wait 03896b2bb2375fcd45efe0f1b64e95d417846dd804da9fe93d76ecf6f19c74b4
  | 2
  | + docker cp 03896b2bb2375fcd45efe0f1b64e95d417846dd804da9fe93d76ecf6f19c74b4:/extract.tar bazel-out/k8-fastbuild/bin/tests/config/debian-jessie-autoconfig_extract.tar
[...]

PR #240, opened 2018-11-16
Fix failing tests resulting from buggy 0.19.1 release

nlopezgi:
Fixes:
https://github.com/bazelbuild/bazel-toolchains/issues/239

Related to https://github.com/bazelbuild/bazel/issues/6665


tests/config/BUILD
```diff
@@ -27,6 +27,9 @@ load(
     "gcc_env",
 )
 
+bazel_latest = "0.19.0"
+bazel_latest_rc = "1"
```
xingao267: This is confusing. What if the latest release is 0.20.0, and the latest rc is 0.21.0 rc2. What do we put in bazel_latest and bazel_latest_rc?

PR #241, opened 2018-11-17
Use container URLs that work externally.

EdSchouten:
When using a stock Docker installation without any credentials set up,
containers under gcr.io/cloud-marketplace cannot be pulled. This is
likely why the RBE Docker container documentation suggests that
launcher.gcr.io is used instead:

https://storage.googleapis.com/rbe-toolchain-docs/getting-started-rbe-debian8.pdf

Using launcher.gcr.io here will make Bazel add these names to the
REv2's Platform message. This allows on-premise builder implementations
(e.g., Buildbarn) to do on-demand container image downloading
without requiring any special configuration of credentials.


xingao267:
Hi @EdSchouten, thanks for you PR. We actually intentionally made the switch from launcher.gcr.io (l.gcr.io) to marketplace.gcr.io (https://github.com/bazelbuild/bazel-toolchains/pull/191) because the launcher.gcr.io (l.gcr.io) is planned to be deprecated in the future by their team (timeline is still undetermined). That [document](https://storage.googleapis.com/rbe-toolchain-docs/getting-started-rbe-debian8.pdf) is out-of-date, and I need to update that.

To pull images from marketplace.gcr.io, you just need to follow the instructions [here](https://cloud.google.com/sdk/gcloud/reference/auth/configure-docker), and the images are available to everyone that is logged in. Thanks again.

EdSchouten:
This means in effect that we won't be providing access to remote build base images for users that don't have a Google account, right?

nlopezgi:
yes, do you see this as an issue for any specific use cases?

EdSchouten:
Just a minor breakage on my side: phasing out `launcher.gcr.io` would break the build of Buildbarn as it's currently defined. I ship two container images based on RBE Debian 8 and Ubuntu 16.04 that include a runner binary for remote builds:

https://github.com/EdSchouten/bazel-buildbarn/blob/89aedb7380d2609b3a2492c5fd05fc5466a46eb2/cmd/bbb_runner/BUILD.bazel#L35-L57
https://github.com/EdSchouten/bazel-buildbarn/blob/89aedb7380d2609b3a2492c5fd05fc5466a46eb2/WORKSPACE#L56-L68

Switching that over to use use `marketplace.gcr.io` shouldn't be hard, but has the downside that building Buildbarn becomes restricted to people having a Google Cloud account.

On a more general note: I think it's a bit odd if:
- Bazel is a community project that can be downloaded without any form of registration,
- the bazel-toolchains repository is part of that community project,
- accessing the container images on which these rules depend does require registration.

If one of the other big cloud providers (Amazon, Microsoft) would provide similar container images that are only accessible to people having an account there, would we also accept pull requests here for toolchain configurations for those images? Or would we ask them to host those under their own GitHub organization?

That said, this doesn't affect me personally. Just want to make sure the right trade-offs are made.

PR #254, opened 2018-12-04
Improve error message of gcs_file & add test

smukherj1:
Make sure we clearly print the name of the GCS file the file was
downloaded from, the expected and actual digests. Also adds a simple
test that ensures the downloaded file actually exists.

The new error message is very helpful when updating GCS paths of files. The new digest is shown in the error and can be used to update the target attributes


tests/rules/BUILD
```diff
@@ -0,0 +1,8 @@
+# Simply ensures the file supposed to be downloaded by the gce_file
+# test rule was actually downloaded
+sh_test(
```
nlopezgi: Why not use a file_test (https://github.com/bazelbuild/bazel/blob/master/tools/build_rules/test_rules.bzl#L306)?

PR #263, opened 2018-12-14
Provide sample WORKSPACE and bazelrc files to use with rbe_autoconf

nlopezgi:
Adding sample files to remotebuildexecution/rbe_system_check.
All the user will need to do now is replace template values in WORKSPACE file and add their stack specific flags.


examples/remotebuildexecution/rbe_system_check/.bazelrc.sample
```diff
@@ -0,0 +1,42 @@
+# Copyright 2016 The Bazel Authors. All rights reserved.
```
xingao267: Could you check if the internal bazelrc validation tests continue to work with this?

nlopezgi: We're switching to a different set of tests that work with this.

PR #269, opened 2018-12-27
Generated new toolchain configs for Bazel 0.21.0

alex1545:
Generated new toolchain configs for Bazel 0.21.0


release/license.tpl
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 The Bazel Authors. All rights reserved.
+# Copyright 2019 The Bazel Authors. All rights reserved.
```
xingao267: Don't change the license year here. It should be the time when Bazel is first launched.

bazelrc/bazel-0.21.0.bazelrc
```diff
@@ -0,0 +1,127 @@
+# Copyright 2019 The Bazel Authors. All rights reserved.
```
xingao267: 2016

configs/debian8_clang/0.4.0/bazel_0.21.0/cpp/BUILD
```diff
@@ -0,0 +1,54 @@
+# Copyright 2019 The Bazel Authors. All rights reserved.
```
xingao267: 2016

release/license.tpl
xingao267: We generally follow the rule of using the year when the product is first launched. It's sometimes not clear whether we should use 2016 or 2017 in this repo. I think we use 2016 for Bazel generated configs and 2017 for rules that we wrote and is specific to this repo (which is launched in 2017). For new repos to be created, we will use 2019.

PR #273, opened 2019-01-09
add configs for ubsan for Bazel 0.21.0

nlopezgi:


xingao267:
Do think we also need to add ubsan configs for Bazel 0.20.0 for completeness?

nlopezgi:
I might add them in an upcoming PR if the ones for 0.21.0 do not work with bazel 0.20.0

PR #280, opened 2019-01-16
Add docker to bazel bases

nlopezgi:


repositories/repositories.bzl
```diff
@@ -120,6 +120,14 @@ def repositories():
             urls = ["https://download.docker.com/linux/ubuntu/gpg"],
         )
 
+    # Docker gpg key necessary to install Docker in the containers.
```
xingao267: It looks like it's the same key for all three. Maybe we can just use one.

nlopezgi: done

PR #281, opened 2019-01-17
only use checked in configs if env, registry and repo are not set

nlopezgi:


rules/rbe_repo.bzl
```diff
@@ -791,6 +789,12 @@ def rbe_autoconfig(
         fail(("Could not find a valid digest for revision %s, " +
               "please make sure it is declared in " +
               "@bazel_toolchains//rules:toolchain_containers.bzl" % revision))
+    # If the user has set a custom env, registry or repository dont use
```
xingao267: don't

PR #282, opened 2019-01-18
Add gcloud to trusty and jessie

nlopezgi:
* also adding explicit downloaded_file_path to gpg targets as images that need more than one key would end up having keys overwritten as the file names were identical


container/debian8/layers/gcloud/BUILD
```diff
@@ -0,0 +1,37 @@
...
+load("//rules/container:docker_toolchains.bzl", "language_tool_layer")
+load(
+    "//container/common/gcloud:version.bzl",
+    "GCLOUD_VERSION",
+)
+
+language_tool_layer(
+    name = "gcloud-ltl",
+    additional_repos = [
+        "deb http://packages.cloud.google.com/apt cloud-sdk-xenial main",
```
xingao267: cloud-sdk-xenial

is this the right one?

nlopezgi: done

PR #288, opened 2019-01-29
DO NOT REVIEW YET. Delete legacy container targets.

xingao267:
These targets have been migrated to either
https://github.com/GoogleCloudPlatform/layer-definitions or
https://github.com/GoogleCloudPlatform/container-definitions.


README.md
```diff
@@ -18,3 +18,8 @@ Release information of toolchain configs can be found at:
 https://releases.bazel.build/bazel-toolchains.html.
 
 This repository also hosts the skylark rule used to generate toolchain configs.
+
+This repo previously contains Bazel targets that are used to generate toolchain
```
nlopezgi: contained

tests/container/BUILD
```diff
@@ -1,120 +0,0 @@
...
-licenses(["notice"])  # Apache 2.0
-
-package(default_visibility = ["//visibility:public"])
-
-load("@base_images_docker//package_managers:download_pkgs.bzl", "download_pkgs")
-load("@base_images_docker//package_managers:install_pkgs.bzl", "install_pkgs")
-load("@base_images_docker//package_managers:apt_key.bzl", "add_apt_key")
-load("//rules/container:docker_toolchains.bzl", "toolchain_container")
-
-# This file contains some sample targets that excersise the container rules
```
nlopezgi: lets leave some of these basic tests for toolchain_container

nlopezgi: Maybe the one that we want to leave is the one currently in tests/config/BUILD (so that it can be used to test toolchain_container and then also test docker_toolchain_autoconfig), I do think if so, we should move the toolchain_container target from that build file (https://github.com/bazelbuild/bazel-toolchains/blob/master/tests/config/BUILD#L77 ) to this one

xingao267: done

PR #291, opened 2019-01-30
Move buildifier to buildkite.

xingao267:


configs/ubuntu16_04_clang/1.0/BUILD
```diff
@@ -155,5 +155,9 @@ platform(
           name: "container-image"
           value:"docker://gcr.io/asci-toolchain/nosla-xenial-docker@sha256:780a96ba94b7666e31ba9a184e5a1742c3398250d47f5b87b1b4a60f8c41f164"
         }
+        properties: {
```
nlopezgi: Do we need this for the 1.0 configs?

xingao267: If we want to use that platform then yes. Although it's not used, but we might as well add the property there as well.

PR #299, opened 2019-02-05
Fix build status badge to show status from master only.

philwo:
Failures in other branches / PRs will otherwise cause it to turn red. :)


philwo:
Sorry - I'll fix the "'ascii' codec" error in the bazelci.py script. :|

PR #316, opened 2019-02-15
Replace --genrule_strategy with --strategy=Genrule

perfinion:
Bazel 0.21.0 and higher have added --strategy=Genrule.
--genrule_strategy does not take effect if --strategy=Genrule is set
causing genrules to not be run remotely.

See-also: https://github.com/bazelbuild/bazel/issues/6760
Signed-off-by: Jason Zaman <jason@perfinion.com>

### DIFF
--- a/bazelrc/bazel-0.21.0.bazelrc
+++ b/bazelrc/bazel-0.21.0.bazelrc
@@ -59,7 +59,7 @@ build:remote --platforms=@bazel_toolchains//configs/ubuntu16_04_clang/1.1:rbe_ub
 build:remote --spawn_strategy=remote
 build:remote --strategy=Javac=remote
 build:remote --strategy=Closure=remote
-build:remote --genrule_strategy=remote
+build:remote --strategy=Genrule=remote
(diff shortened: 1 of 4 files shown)


perfinion:
@nlopezgi

ulfjack:
The downside of this change is that it doesn't work with earlier Bazel versions.

perfinion:
https://docs.bazel.build/versions/0.21.0/command-line-reference.html This says 0.21.0 has --strategy=Genrule so I only changed the bazelrc files for 0.21 and 0.22. 
I could un-do the changes in release/ and example/ if you'd like.

nlopezgi:
thanks @perfinion for creating this PR and thanks @ulfjack for your comment. We will release the change of flags with the bazelrc for 0.23.0 as --genrule_strategy is still supported with 0.21.0 and 0.22.0 and making this change for 0.21.0 and 0.22.0 properly would involve re-releasing these configs (https://releases.bazel.build/bazel-toolchains.html)

PR #326, opened 2019-02-28
Make rbe_default to use digest instead of tag.

xingao267:


xingao267:
I'll cut the release after this fix.

PR #322, opened 2019-02-27
Support using checked-in configs with containers that extend rbe-ubuntu

nlopezgi:
NOT READY FOR REVIEW.
This PR is blocked for review on https://github.com/bazelbuild/bazel-toolchains/pull/327
Will change message once its ready to review.

This PR also includes:

- Fix RBE autoconf so that directory for configs is properly created.
- Changing data structure for storing rbe-ubuntu shas to config directories (all data moved to configs/ubuntu16_04_clang/versions.bzl)
  - Use clang version instead of legacy 1.0, 1.1 versioning
- rbe_autoconfig now always generates a platform (i.e., not an alias when using checked in configs)
   - When using checked in configs only the cc_toolchain + java_runtime are aliases
- Changed slightly structure of produced artifacts (cc directory for cc toolchain targets, java directory for java toolchain targets, config directory for platform targets) 
  - Added new buildkite_config target in WORKSPACE
      - Use nosla-ubuntu16_04-bazel-docker-gcloud (as its being kept up to date better than the one we were using)

MOVED TO SEPARATE PR:
moved to https://github.com/bazelbuild/bazel-toolchains/pull/328:
- Switch BuildKite CI to use rbe autoconfig for this repo 

moved to https://github.com/bazelbuild/bazel-toolchains/pull/327:
- Generating configs to use with RBE autoconf (created in directories corresponding to clang version)
   - Code under configs/ubuntu16_04_clang/8.0.0/ was all generated with rbe_autoconfig using released container (0.20-0.23)
   - Code under configs/ubuntu16_04_clang/9.0.0/ was all generated with rbe_autoconfig using staging container (0.23 only).

PR #327, opened 2019-03-01
adding generated configs with new rbe_Autoconf

nlopezgi:
This change is splitting of the auto-generated files from https://github.com/bazelbuild/bazel-toolchains/pull/322.
Lets review this one first (its only generated files) so that the other one gets easier to review.


xingao267:
Did you verify that these are the same as those in the repo before?

nlopezgi:
> Did you verify that these are the same as those in the repo before?

The configs are being tested in GoB repo CL (switching all tests to use this new rule) and in PR https://github.com/bazelbuild/bazel-toolchains/pull/328 (testing new feature to use checked in configs with container that extends rbe ubuntu16_04)

PR #329, opened 2019-03-01
Test toolchain issue

nlopezgi:
Do not review. 
Just sending this with testing changes to verify if the issue that I had yesterday related to error messaging for toolchains can be reproduced easily. ref: https://github.com/bazelbuild/bazel/issues/7597


nlopezgi:
For record:
CI run with failure "errors encountered while analyzing target":
https://buildkite.com/bazel/bazel-toolchains/builds/1338#71adf950-db86-4fdf-a20d-3847d6b436a0

CI run in which build starts:
https://buildkite.com/bazel/bazel-toolchains/builds/1339#6a57c93c-0194-42e4-9a40-4058ff6cb1ff

The only difference between these two is that the `toolchain` target has a broken `toolchain` attr in the broken case vs the passing case (https://github.com/bazelbuild/bazel-toolchains/pull/329/commits/dc367cf78a17d9df134251eed44390e661e9fd30).

The failed run does not show any error messages similar to the one mentioned by @katre in https://github.com/bazelbuild/bazel/issues/7597#issuecomment-468712033

PR #344, opened 2019-03-06
preparing bazelrc files for launch of rbe_autoconfig

nlopezgi:
The bazelrc file will replace the existing latest.bazelrc
created contrib.bazelrc to move content that has a lower SLA (i.e., not actively being tested / maintained)


bazelrc/contrib.bazelrc
```diff
@@ -0,0 +1,47 @@
+# Copyright 2016 The Bazel Authors. All rights reserved.
```
xingao267: I'm not sure if we want this file?
What's the plan for the current bazelrc with Bazel versions and the latest.bazelrc?

nlopezgi: spoken offline, the references to latest.bazelrc for local docker sandbox or remote cache will be changed to point to this file. The latest.bazelrc files should be deprecated and even removed from our repo soon.

PR #345, opened 2019-03-06
Fix buildifier 0.22.0 lint issues & disable buildifier

smukherj1:


fweikert:
Note that https://github.com/bazelbuild/continuous-integration/pull/542 (still pending) will allow you to set the Buildifier version in the configuration. Moreover, "buildifier: true" won't be valid any more (it'll be "buildifier: latest").

smukherj1:
@fweikert We are waiting for https://github.com/bazelbuild/continuous-integration/issues/531 to be fixed because we to turn on most of the new lint warnings in 0.22.0 which we worked to fix here but turn off a few specific warnings that are potentially breaking.

fweikert:
I already have a fix for #531 - I'll ping this thread once it's merged and deployed.

fweikert:
Fix has been deployed. Documentation: https://github.com/bazelbuild/continuous-integration/tree/master/buildkite#running-buildifier-on-ci

nlopezgi:
thanks @fweikert !

PR #360, opened 2019-03-12
Make java_home optional in rbe_autoconfig

smukherj1:
Not specifying java_home and specifying a container with a blank
java_home or explicitly specifying a blank java_home attribute results
in skipping over java config generation.

### DIFF
--- a/rules/rbe_repo.bzl
+++ b/rules/rbe_repo.bzl
@@ -281,7 +281,8 @@ def _impl(ctx):
     # Get the value of JAVA_HOME to set in the produced
     # java_runtime
     java_home = _get_java_home(ctx, image_name)
-    _create_java_runtime(ctx, java_home)
+    if java_home != "":
+        _create_java_runtime(ctx, java_home)
 
     # run the container and extract the autoconf directory
     _run_and_extract(
@@ -299,6 +300,7 @@ def _impl(ctx):
         ctx,
         bazel_version = ctx.attr.bazel_version,
         project_root = project_root,
+        copy_java = java_home != ""
     )
 
 # Convenience method to print results of execute (and fail on errors if needed).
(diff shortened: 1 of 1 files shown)


rules/rbe_repo.bzl
```diff
@@ -589,9 +588,16 @@ def _create_platform(ctx, image_name, name):
...
-def _expand_outputs(ctx, bazel_version, project_root):
+def _expand_outputs(ctx, bazel_version, project_root, copy_java):
+    """
+    Copies all outputs of the autoconfig rule to a directory in the project.
+
+    Args:
+        ctx: The Bazel context.
+        bazel_version: The Bazel version string.
+        project_root: The output directory where configs will be copied to.
+        copy_java: Boolean indicating whether java configs should be copied.
```
nlopezgi: just pass down the boolean attr (or read it from ctx in this def)

rules/rbe_repo.bzl
```diff
@@ -689,8 +697,8 @@ _rbe_autoconfig = repository_rule(
             doc = ("Optional. The location of java_home in the container. " +
                    "For example, '/usr/lib/jvm/java-8-openjdk-amd64'. If " +
                    "not set, the rule will attempt to read the JAVA_HOME env " +
-                   "var from the container. If that is not set the rule will " +
-                   "fail."),
+                   "var from the container. If set to a blank string, the " +
+                   "rule will skip java config generation."),
```
nlopezgi: I think we need to add a new boolean attr (as you also mentioned on chat). Particularly because this would not have any equivalent for the cc config. So yes, lets add java_config (and later on cc_config) as boolean attr (or some other name if you have a suggestion). Default value set to true. If set to false, and java_home is not None, error out.

rules/rbe_repo.bzl
smukherj1: Done

rules/rbe_repo.bzl
smukherj1: Done

rules/rbe_repo.bzl
```diff
@@ -846,6 +868,7 @@ def rbe_autoconfig(
         bazel_rc_version = bazel_rc_version,
         digest = digest,
         env = env,
+        create_java_configs = create_java_configs,
```
nlopezgi: can you add a check above (around lines 828-862) that fails if create_java_configs is false and java_home was set

smukherj1: done

PR #361, opened 2019-03-13
Add release commit list for Ubuntu 16.04 Clang container

smukherj1:


configs/ubuntu16_04_clang/releases.yaml
```diff
@@ -0,0 +1,66 @@
+- commitSHA: bccee4855c049d34bac481083b4c68e2fab8cc50
```
xingao267: Add a note that this is an autogenerated file and should not be modified directly.

smukherj1: Done

PR #380, opened 2019-03-27
Add option to copy resources instead of mounting in rbe_autoconfig().

xingao267:
This is needed to use rbe_autoconfig() on CircleCI for Angular project.
https://circleci.com/docs/2.0/building-docker-images/#mounting-folders

### DIFF
--- a/rules/rbe_repo.bzl
+++ b/rules/rbe_repo.bzl
@@ -705,7 +705,9 @@ _rbe_autoconfig = repository_rule(
             doc = (
                 "Optional. Specifies whether to copy instead of mounting " +
                 "resources such as scripts and project source code to the " +
-                "container for Bazel autoconfig."
+                "container for Bazel autoconfig. Note that copy is more " +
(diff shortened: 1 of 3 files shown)


rules/rbe_repo.bzl
```diff
@@ -686,6 +700,14 @@ _rbe_autoconfig = repository_rule(
                    "Bazel version. " +
                    "Used internally when use_checked_in_confs is true."),
         ),
+        "copy_resources": attr.bool(
+            default = False,
+            doc = (
+                "Optional. Specifies whether to copy instead of mounting " +
+                "resources such as scripts and project source code to the " +
+                "container for Bazel autoconfig."
```
nlopezgi: Note that copy is more expensive and should only be enabled where mounting is not enabled

xingao267:
So far only tested with Angular project. Will add new tests in GoB after it is merged.

xingao267:
Tested internally.

PR #385, opened 2019-03-29
Setup the rules_docker dependency so that it can be updated by copybara

smukherj1:


repositories/repositories.bzl
```diff
@@ -23,18 +23,19 @@ load(
...
+load("@bazel_toolchains//deps:io_bazel_rules_docker.bzl", io_bazel_rules_docker_version = "version")
 
 def repositories():
     """Download dependencies of bazel-toolchains."""
     excludes = native.existing_rules().keys()
 
     # ============================== Repositories ==============================
     if "io_bazel_rules_docker" not in excludes:
-        http_archive(
+        git_repository(
```
xingao267: Can we add a TODO or note that we should add the SHA check back eventually?

smukherj1: Done

PR #405, opened 2019-04-03
Update to new repo macro for latest rules_go

smukherj1:
This is a manual version of #404 because the latest rules_go version
changes the location of the deps macro from //go:def.bzl to
//go:deps.bzl.

IO_BAZEL_RULES_GO_REV_ID: f900a9e520e9fdb7b81511f9aa5de2e43da19e2a

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -49,7 +49,7 @@ load(
 
 bazel_toolchains_images()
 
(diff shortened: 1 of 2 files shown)


xingao267:
Add a comment about why this is not handled by the automatic job.

PR #410, opened 2019-04-05
Add a GCB config to test toolchain configs on RBE

smukherj1:


cloudbuild.yaml
```diff
@@ -0,0 +1,11 @@
+# Launch a RBE hello world build + unit test using the @rbe_default toolchain
+# configs.
+steps:
+- name: 'gcr.io/cloud-builders/bazel'
```
nlopezgi: use our bazel container

smukherj1: done

PR #416, opened 2019-04-08
Rbe autoconf tests

nlopezgi:
Basic unit tests that verify the file outputs in the external repo folder for now. Will create more complex unit tests in upcoming PR.
Setting parallelism of analysis phase that runs these tests to 1 as rbe_autoconfig targets cannot run in parallel.

Tests work as follows:
* As part of rbe_autoconfig rule creates (optionally) a copy of all generated artifacts (BUILD files, bzl files, etc). BUILD files are renamed to test.BUILD (to avoid Bazel reading them). Rule also creates a root BUILD file in test copy that has a single filegroup. This enables all tests to refer to these tests artifacts properly.
* Tests then declare dependency on test copy filegroup and can make assertions
* This first PR adds basic tests that verify files are present according to basic attrs of rule (whether checked in or generated configs are used, whether java/cc configs are generated).
* Will be adding additional tests in upcoming PRs
* Note generated config tests are not run in BuildKite CI as it does not support docker yet.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -192,22 +192,22 @@ rbe_autoconfig(
 rbe_autoconfig(
     name = "rac_generate",
     bazel_version = _ubuntu1604_bazel,
-    use_checked_in_confs = False,
     create_testdata = True,
+    use_checked_in_confs = False,
 )
 
 rbe_autoconfig(
     name = "rac_generate_no_java",
     bazel_version = _ubuntu1604_bazel,
     create_java_configs = False,
-    use_checked_in_confs = False,
     create_testdata = True,
+    use_checked_in_confs = False,
 )
 
 rbe_autoconfig(
     name = "rac_generate_no_cc",
     bazel_version = _ubuntu1604_bazel,
     create_cc_configs = False,
-    use_checked_in_confs = False,
     create_testdata = True,
+    use_checked_in_confs = False,
 )
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -118,7 +118,9 @@ rbe_autoconfig(
 rbe_autoconfig(
     name = "rbe_default_with_output_base",
     config_dir = "default",
-    output_base = "configs/ubuntu16_04_clang/1.1",
+    create_cc_configs = False,
+    external_repos = ["docker_config"],
+    output_base = "tests/temp",
 )
 
 # Targets used by automatic config generation and release service.
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -170,34 +170,34 @@ rbe_autoconfig(
 # Targets below for purposes of testing of rbe_autoconfig rule only
 
 rbe_autoconfig(
-    name = "rac_checked_in",
+    name = "rbe_autoconf_checked_in",
     bazel_version = _ubuntu1604_bazel,
     create_testdata = True,
 )
 
 rbe_autoconfig(
-    name = "rac_checked_in_no_java",
+    name = "rbe_autoconf_checked_in_no_java",
     bazel_version = _ubuntu1604_bazel,
     create_java_configs = False,
     create_testdata = True,
 )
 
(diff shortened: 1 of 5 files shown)


rules/rbe_repo.bzl
```diff
@@ -692,6 +711,33 @@ def _expand_outputs(ctx, bazel_version, project_root):
...
+    # create a root BUILD file with a filegroup
+    ctx.file("test/BUILD", """package(default_visibility = ["//visibility:public"])
+exports_files(["empty"])
+filegroup(
+    name = "exported_testdata",
+    srcs = glob(["**/*"]),
+)
+""", False)
+
+    # create an empty file to reference in tests
```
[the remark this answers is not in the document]
nlopezgi: done

WORKSPACE
```diff
@@ -179,3 +166,48 @@ rbe_autoconfig(
     registry = "gcr.io",
     repository = "asci-toolchain/nosla-ubuntu16_04-bazel-docker-gcloud",
 )
+
+# Targets below for purposes of testing of rbe_autoconfig rule only
+
+rbe_autoconfig(
+    name = "rac_checked_in",
```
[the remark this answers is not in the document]
nlopezgi: replaced for rbe_autoconf

WORKSPACE
```diff
@@ -103,24 +103,11 @@ gcs_file(
 
 load("@bazel_toolchains//rules:rbe_repo.bzl", "rbe_autoconfig")
 
-rbe_autoconfig(name = "rbe_default")
```
smukherj1: Update //bazelrc/.bazelrc to ensure it keeps working with this change? I intend to use it in my GCB job to test rbe_autoconfig

rules/rbe_repo.bzl
```diff
@@ -317,6 +321,10 @@ def _impl(ctx):
         project_root = project_root,
     )
 
+    # Copy all outputs to the test directory
+    if ctx.attr.create_testdata:
```
smukherj1: My feeling is this calls for a refactor where the current body of the impl function goes to an inner "generate_configs" function and the _copy_to_test_dir(ctx) function is called when that returns. This way we don't risk breaking the delicate logic that requires _copy_to_test_dir(ctx) is called right before returning if additional return statements are added in a refactor.

rules/rbe_repo.bzl
```diff
@@ -692,6 +711,35 @@ def _expand_outputs(ctx, bazel_version, project_root):
...
+    # create a root BUILD file with a filegroup
+    ctx.file("test/BUILD", """package(default_visibility = ["//visibility:public"])
+exports_files(["empty"])
+filegroup(
+    name = "exported_testdata",
+    srcs = glob(["**/*"]),
+)
+""", False)
+
+    # Create an empty file to reference in test.
```
smukherj1: Technically, I don't think this is necessary because doing "$(location @rbe_autoconf_generate_no_cc//test:exported_testdata)" should work as well. When the file group has more than one file, this should just randomly pick one file from that group. Since you are only using this to obtain the root directory, that should serve your purpose.

WORKSPACE
nlopezgi: re-added the target.

rules/rbe_repo.bzl
nlopezgi: I dont want to run into issues with the functionality of $(location ...) for a filegroup, its more reliable to just create the file.

rules/rbe_repo.bzl
nlopezgi: I agree that a refactor would be desirable, given that we only have 2 paths I'll delay doing the costly refactor until we have more paths. Also, it should be fairly easy to notice if changes don't respect the current invariant (e.g., tests would break), added a TODO.

PR #419, opened 2019-04-09
Fix GCB CI job so that it can successfully launch RBE builds

smukherj1:
1. Make the GCB job launch RBE builds to the RBE instance on the GCP
   project hosting the GCB job.
2. Upload logs to a bucket owned by toolchains team to ensure toolchains
   team has the ability to set the ACL permissions on the logs.


smukherj1:
Don't review or merge yet. GCB presubmit job isn't running for some reason. I'm investigating

PR #422, opened 2019-04-09
if config_repos or copy_resources dont use checked in configs

nlopezgi:


nlopezgi:
closing as its not probably the right thing to do to just generate configs always when these options are used.

PR #454, opened 2019-04-18
improve validations in rbe_autoconfig

nlopezgi:
* Add validations to check when docker image will be needed to be pulled
* Split off logic to resolve project_root from main impl method
* Produce a dummy file if an image was pulled (for tests)
* Add tests that validate an image will not be pulled in edge case
* Add tests to valiate image was pulled in other cases

### DIFF
--- a/rules/rbe_repo.bzl
+++ b/rules/rbe_repo.bzl
@@ -230,39 +230,15 @@ def _rbe_autoconfig_impl(ctx):
     # Use l.gcr.io registry to pull marketplace.gcr.io images to avoid auth
     # issues for users who do not do gcloud login.
     image_name = image_name.replace("marketplace.gcr.io", "l.gcr.io")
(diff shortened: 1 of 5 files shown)


rules/rbe_repo.bzl
```diff
@@ -416,6 +436,10 @@ def _pull_image(ctx, docker_tool_path, image_name):
     _print_exec_results("pull image", result, fail_on_error = True)
     ctx.report_progress("image pulled.")
 
+    # Create a dummy file with the image name to enable testing
+    # if container was pulled
+    ctx.file("image_name", image_name, False)
```
xingao267: Write some message to the file to indicate this file is just to indicating if the container is pulled or not?

nlopezgi: done

PR #473, opened 2019-04-29
enable force using checked in configs + RBE on GCB tests + various fixes

nlopezgi:
* Adding attr that enables failing if rbe_autoconfig was not able to use checked-in configs (as requested in toolchain config launch review)

* delete outputs in external repo when output_base is used: its not possible to sanely produce artifacts that will be usable both from the @external_repo/.. and //output_base/.... So just eliminate them if output_base is used

* adding more tests that run RBE on GCB using the configs in this repo + rbe_autoconfig + RBE.

* stop using rsync (its not present in bazel containers and really not needed)


rules/rbe_repo.bzl
```diff
@@ -1077,6 +1102,11 @@ def rbe_autoconfig(
         use_checked_in_confs = use_checked_in_confs,
     )
 
+    if force_checked_in_confs and not config_version:
```
smukherj1: Consider emitting an error if use_checked_in_confs is set to False when force_checked_in_confs is set to True

WORKSPACE
```diff
@@ -103,7 +103,10 @@ gcs_file(
 
 load("@bazel_toolchains//rules:rbe_repo.bzl", "rbe_autoconfig", "rbe_autoconfig_root")
 
-rbe_autoconfig(name = "rbe_default")
+rbe_autoconfig(
+    name = "rbe_default",
+    force_checked_in_confs = True,
```
xingao267: What about merging two flags: "force_checked_in_confs" and "use_checked_in_confs" to be just one flag, maybe "use_checked_in_confs", and the enum value can be "force" (error out if no checked in configs are available), "attempt" (try finding checked-in configs and fall back to generate configs on the fly), and "false" (always generate the configs on the fly).

tests/rbe_repo/generate_confs.yaml
```diff
@@ -0,0 +1,101 @@
+# Steps in this file use toolchain configs produced by rbe_autoconf**
```
xingao267: Add some individual comments for each RBE build about what it's testing?

WORKSPACE
nlopezgi: done

tests/rbe_repo/generate_confs.yaml
nlopezgi: done

rules/rbe_repo.bzl
nlopezgi: moved to using a single attr for both

PR #496, opened 2019-05-10
update skylib and fix issue with test depending on skylib

nlopezgi:
I had created a test depending on building @bazel_skylib, which is not really a good idea to begin with. Latest skylib update breaks the test. Removing that dependency. I really like how our bots updating deps makes it easier to catch these issues early.

replaces https://github.com/bazelbuild/bazel-toolchains/pull/494


nlopezgi:
pending trigger was successful but for some reason did not report back

PR #509, opened 2019-05-15
add new config + bazelrc file to test incompatible flags

nlopezgi:


bazelrc/.bazelrc.test.0.25.0
```diff
@@ -0,0 +1,46 @@
...
+build:remote --define=EXECUTOR=remote
+build:remote --remote_cache=remotebuildexecution.googleapis.com
+build:remote --remote_executor=remotebuildexecution.googleapis.com
+build:remote --tls_enabled=true
+build:remote --remote_timeout=3600
+build:remote --auth_enabled=true
+
+# Turn on incompatible_enable_cc_toolchain_resolution and don't
+# use crosstool_top
+build:remote --incompatible_enable_cc_toolchain_resolution
```
xingao267: Why not use the existing bazelrc and use this flag in the test?

nlopezgi: The default bazelrc flag has action_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1 which, as I found out doing this test, is incompatible with use of incompatible_enable_cc_toolchain_resolution.

Also, I want to define a way that we can do tests for different versions of Bazel with different incompatible  flags, changing only the minimum set of artifacts for a new Bazel release (i.e., not just for this specific incompatible_enable_cc_toolchain_resolution flag). This way all you will need to add for a new version of bazel (and new set of incompatible flags) is the bazelrc file and the new trigger.

nlopezgi: Moved all files to yaml file and removed custom bazelrc file

nlopezgi:
Pending trigger succeeded but did not report back success. This is ready now for review again.

PR #526, opened 2019-05-23
Support not using bazel-toolchains as source for pre-checked in configs

nlopezgi:
High level features added in this PR:
* decouple rbe_autoconfig from bazel-toolchains repo
* enable use of rbe_autoconfig to generate / pick configs from an arbitrary repo (e.g., for complex use cases where users have possibly multiple custom containers and a github repo to check in all their configs).

More features (smaller)
* better matching of potential checked in configs to use
* better error messaging of why checked in config was not picked
* versions.bzl file (used by rbe_autoconf to figure out checked-in confs) is now maintained/updated automatically 
* single rbe_autoconfig target can be used both to generate configs and to use checked-in configs

- Migration pending items:
- [ ] BEFORE PR IS IN: Modify the toolchain config service to add new declarations in versions.bzl file
- [ ] AFTER PR IS IN: Turn down generation of versions.bzl file and switch to new rbe_default

PR #535, opened 2019-05-29
Create rbe_repo definitions and validations

nlopezgi:
This PR splits off a part of https://github.com/bazelbuild/bazel-toolchains/pull/526
that can be discussed, reviewed and merged independently of that large PR.

This is just adding the new structures that are used in https://github.com/bazelbuild/bazel-toolchains/pull/526 + a basic test from rbe_repo.bzl that validates the default rbe_repo
is properly constructed.


rules/rbe_repo/rbe_repo_spec.bzl
```diff
@@ -0,0 +1,435 @@
...
+    _validate_default_toolchain_config_spec(name, rbe_repo_spec)
+    _validate_bazel_to_config_spec_names_map(name, rbe_repo_spec)
+    _validate_container_to_config_spec_names_map(name, rbe_repo_spec)
+    _validate_configs(name, rbe_repo_spec)
+
+def _validate_default_toolchain_config_spec(name, rbe_repo_spec):
+    if (rbe_repo_spec["rbe_repo_gen_spec"].default_toolchain_config_spec != "" and
+        type(rbe_repo_spec["rbe_repo_gen_spec"].default_toolchain_config_spec) != "struct"):
+        fail(("%s has a rbe_repo_spec[\"rbe_repo_gen_spec\"] field 'default_toolchain_config_spec' " +
+              "that is not either an empty string or a 'struct'. " +
```
smukherj1: Suggest using neither nor instead of "not either"

rules/rbe_repo/rbe_repo_spec.bzl
```diff
@@ -0,0 +1,435 @@
...
+    Args:
+      name: Name of the rbe_autoconfig rule
+      rbe_repo_spec: The rbe_repo_spec to validate
+
+    """
+    _validate_repo_spec_keys(name, rbe_repo_spec)
+    _validate_rbe_repo_gen_spec(name, rbe_repo_spec)
+
+# Validates the rbe_repo_spec is a struct and has all required top level keys
+def _validate_repo_spec_keys(name, rbe_repo_spec):
```
smukherj1: Quite a bit of repetition in these functions. I wonder if this could be refactored into yet another "type schema" that maps from expected attribute names to their types and uses that to specify the expected type for the various structs being validated, marshaled & unmarshaled.

So for example a type schema could be
```
config_repo_type_schema = {
"bazel_to_config_spec_names_map": "dict",
 ...
}
```

Also, all this parsing and validation appears too heavyweight for Starlark in my opinion. Should we consider creating these configurations in other format like YAML, etc and have a utility in Go/Python etc do the high level parsing and validation and generate Starlark files with the processed & validated data. This way we avoid bending over backwards trying to work with Starlark's limitations when doing complex operations like schema validation. We generated Starlark files would then be available for import to other rules that want to use them. Doesn't have to be part of this PR but something to think about.

tests/rbe_repo/BUILD
```diff
@@ -278,10 +278,13 @@ file_test(
 
 # Tests validate run_and_extract uses mount volumes (-v) vs
 # data volumes.
+# Set with tag 'manual' as this test messes up others if the
```
smukherj1: Could you briefly mention here what aspect of the test results in this issue?

rules/rbe_repo/rbe_repo_spec.bzl
nlopezgi: managed to reduce some of the repetition. ptal.
wrt doing this in anything other than starlark, it would be really hard:
- If the data is an attr of the rule we have to process it in the analysis phase which will have to be starlark
- If we move this to outside of attrs that are passed to the repo rule, then changes to the underlying file will not trigger rebuilds as this is a repo rule and deps cannot be really expressed.

rules/rbe_repo/rbe_repo_spec.bzl
nlopezgi: fixed

tests/rbe_repo/BUILD
nlopezgi: added more details. ptal

rules/rbe_repo/toolchain_config_suite_spec.bzl
```diff
@@ -0,0 +1,461 @@
...
+            config_repos = ctx.attr.config_repos,
+            env = ctx.attr.env,
+        )]
+    return configs
+
+# Fail if the config defined by the user in the ctx/selected by default
+# does not match details of the one in versions.bzl
+def _check_config(ctx, config):
+    if not ctx.attr.env == config.env:
+        fail(("%s failed. '%s' was passed as env and '%s' was " +
```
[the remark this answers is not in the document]
nlopezgi: replaced for !=

PR #539, opened 2019-05-30
fix output_base test

nlopezgi:
Missing change from my previous PR that caused a non blocking GCB run to fail (looks like GCB tries to get smart with bazel-* dirs).


nlopezgi:
pending triggers have already passed but did not report success back. This is ready for review

PR #541, opened 2019-05-31
add OSFamily Linux to generated platforms

nlopezgi:
Platforms produced by rbe_autoconfig will now have a new property
properties {
    name: "OSFamily"
    value:  "Linux"
  }
This change should not affect users that only build on linux.
Change is only effective for platforms generated after this PR, we will not be amending old platforms to add this property. To use this, you must migrate to rbe_autoconfig.


nlopezgi:
readding lgtm as this was just a branch update

PR #545, opened 2019-06-03
Add pip dependencies of base images docker

smukherj1:
This updates base images docker to the same version as #527 with fixes
for the build failures.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -420,3 +420,16 @@ load("//rules/rbe_repo:util.bzl", "rbe_autoconfig_root")
 # Needed for testing purposes. Creates a file that exposes
 # the value of RBE_AUTOCONF_ROOT
 rbe_autoconfig_root(name = "rbe_autoconfig_root")
+
+# Pull in dependencies of base_images_docker.
+load("@base_images_docker//package_managers:repositories.bzl", package_manager_deps = "deps")
+package_manager_deps()
+load("@io_bazel_rules_python//python:pip.bzl", "pip_import", "pip_repositories")
+pip_repositories()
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -420,3 +420,21 @@ load("//rules/rbe_repo:util.bzl", "rbe_autoconfig_root")
 # Needed for testing purposes. Creates a file that exposes
 # the value of RBE_AUTOCONF_ROOT
 rbe_autoconfig_root(name = "rbe_autoconfig_root")
+
+# Pull in dependencies of base_images_docker.
+load("@base_images_docker//package_managers:repositories.bzl", package_manager_deps = "deps")
+
+package_manager_deps()
+
+load("@io_bazel_rules_python//python:pip.bzl", "pip_import", "pip_repositories")
```
nlopezgi: lets try to move all this to https://github.com/GoogleContainerTools/base-images-docker/blob/master/package_managers/repositories.bzl

smukherj1: Mentioning why this wasn't done in a top level comment for visibility.

smukherj1:
Response to review [comment](https://github.com/bazelbuild/bazel-toolchains/pull/545#discussion_r289962728)

`load` statements in Starlark may not be called from the body of a macro. In order to import pip dependencies, we must first import `rules_python`, load the pip rules from `rules_python` and then call the pip rules. Thus, in order for this to work, we would need to split off the `repositories.bzl` in `base-images-docker` into two where the first file imports rules_python and the second imports the pip rules. This complexity wasn't deemed worth the effort and was abandoned.

PR #549, opened 2019-06-05
add tests that run RBE builds with external projects

nlopezgi:
Tests run as GCB builds.
- Adding test that run with latest bazel with latest bazerc file for abseil and bazel

- Also added test that can run with arbitrary bazel version (including rc
versions as it uses basilisk) and arbitrary bazelrc file (i.e., does
not need to match bazel version)

- Also added a 0.27 bazelrc file so we can start testing early
(particularly due to https://github.com/bazelbuild/bazel/issues/8469)

### DIFF
--- a/tests/rbe_external_project/bazel.yaml
+++ b/tests/rbe_external_project/bazel.yaml
@@ -1,0 +1,37 @@
+# Basic test that builds Bazel on RBE using the latest Bazel version
+steps:
+# pull the abseil-cpp project
+- name: 'gcr.io/cloud-builders/git'
+  args:
+  - clone
+  - https://github.com/bazelbuild/bazel.git
+  - external-src
+
+# modify the WORKSPACE file & delete the existing .bazelrc file
+- name: 'gcr.io/gcp-runtimes/ubuntu_16_0_4'
+  args:
+  - ../tests/rbe_external_project/add_rbe_default.sh
+  dir: 'external-src'
+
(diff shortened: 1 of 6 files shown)


tests/rbe_external_project/bazel.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - -//src/test/java/com/google/devtools/build/lib/shell:CommandUsingLinuxSandboxTest
+  - -//src/test/java/com/google/devtools/build/lib:vfs_inmemoryfs_test
+  - -//src/test/java/com/google/devtools/build/lib/blackbox/tests:PythonBlackBoxTest
+  - -//src/test/java/com/google/devtools/build/lib/rules/objc:ObjcRulesTests
+  - -//src/test/java/com/google/devtools/build/lib:worker-tests
+  - -//src/test/java/com/google/devtools/build/lib/buildeventstream/transports:BuildEventTransportTest
+  - -//src/test/java/com/google/devtools/build/lib/blackbox/tests/workspace:BazelEmbeddedSkylarkBlackBoxTest
+  - -//src/test/java/com/google/devtools/build/lib/blackbox/tests/workspace:WorkspaceBlackBoxTest
+  dir: 'external-src'
+timeout: 1200s
```
smukherj1: Max supported timeout on GCB is 10mins I believe.

nlopezgi: that's the default, this is how its extended

PR #558, opened 2019-06-12
add test that uses jdk11

nlopezgi:
* Add detect_java_home as attr to rbe_autoconfig to have users explicitly state if they want java_home to be detected
* add default_java_home attr to toolchain_config_suite_spec to optionally have a fallback if user does not provide one and configs do not exist
* do not validate java for selection of checked-in configs. Since java toolchain can be produced w/o container (if no default_java_home is selected) then no need to use checked-in configs for this. 
* add unit tests for use of detect_java_home
* add basic GCB test that uses jdk11

Pending for upcoming PR: use jdk11 in RBE test that builds Bazel


rules/rbe_repo.bzl
```diff
@@ -845,6 +856,9 @@ def rbe_autoconfig(
       digest: Optional. The digest of the image to pull.
           Should not be set if 'tag' is used.
           Must be set together with 'registry' and 'repository'.
+      detect_java_home: Optional. Default False. Should only be set
```
xingao267: If java_home is not specified, and detect_java_home is false, then the java_home in java_runtime() will be empty? or the java config will not be create at all?

rules/rbe_repo.bzl
```diff
@@ -898,8 +910,11 @@ def rbe_autoconfig(
     if bazel_rc_version and not bazel_version:
         fail("bazel_rc_version can only be used with bazel_version.")
 
-    if not create_java_configs and java_home != None:
-        fail("java_home should not be set when create_java_configs is false.")
+    if not create_java_configs and (java_home != None or detect_java_home):
```
xingao267: if not create_java_configs and (java_home or detect_java_home):

rules/rbe_repo.bzl
```diff
@@ -898,8 +910,11 @@ def rbe_autoconfig(
...
+        fail("java_home / detect_java_home should not be set when " +
+             "create_java_configs is False.")
+    if java_home != None and detect_java_home:
```
xingao267: if java_home and detect_java_home

rules/rbe_repo.bzl
xingao267: Saw the fallback logic below. But that's kinda complicated...

rules/rbe_repo.bzl
```diff
@@ -943,6 +958,28 @@ def rbe_autoconfig(
         use_checked_in_confs = use_checked_in_confs,
     )
 
+    # If create_java_configs was requested but no java_home or detect_java_home was
+    # set, we try to resolve a java_home
+    if create_java_configs and not java_home and not detect_java_home:
+        # If a spec was found and that has a java_home, use it
+        if toolchain_config_spec and toolchain_config_spec.create_cc_configs:
```
xingao267: toolchain_config_spec.create_java_configs?

rules/rbe_repo.bzl
```diff
@@ -943,6 +958,28 @@ def rbe_autoconfig(
...
+            # Fallback to try to using the default_java_home set in the
+            # toolchain_config_suite_spec
+            java_home = toolchain_config_suite_spec.get("default_java_home")
+
+        elif use_checked_in_confs != CHECKED_IN_CONFS_FORCE:
+            # Fallback to detecting the java_home if CHECKED_IN_CONFS_FORCE
+            # was not passed
+            detect_java_home = True
+
+        elif toolchain_config_spec and use_checked_in_confs == CHECKED_IN_CONFS_FORCE:
```
xingao267: Not fully understand this part.

rules/rbe_repo.bzl
nlopezgi: fixed

rules/rbe_repo.bzl
nlopezgi: fixed

rules/rbe_repo.bzl
nlopezgi: fixed

rules/rbe_repo.bzl
nlopezgi: changed to a fail early and documented better, thanks for noting

rules/rbe_repo.bzl
nlopezgi: added notes about fallback logic. ptal

PR #561, opened 2019-06-13
use external project for bazel legacy tests

nlopezgi:
Legacy tests are failing with newer version of rules_go as our deps in the bazel-toolchains repo realistically cannot be executed with older versions of Bazel. Switching to using the external repo that has a simple C target only.

Should unblock https://github.com/bazelbuild/bazel-toolchains/pull/560

### DIFF
--- a/tests/rbe_repo/bazel_legacy_rbe_default.yaml
+++ b/tests/rbe_repo/bazel_legacy_rbe_default.yaml
@@ -16,7 +16,7 @@ steps:
 # that verifies that @rbe_default works for Bazel 0.20.0 - 0.25.0
 # These tests are separate from bazel_tests_cloudbuild.yaml as those
 # tests depend on the bazelrc files numbered by versions.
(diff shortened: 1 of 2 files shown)


tests/rbe_repo/bazel_legacy_rbe_default.yaml
```diff
@@ -16,14 +16,15 @@ steps:
 # that verifies that @rbe_default works for Bazel 0.20.0 - 0.25.0
 # These tests are separate from bazel_tests_cloudbuild.yaml as those
 # tests depend on the bazelrc files numbered by versions.
-# However, tests covered by this file have a bazerc file that did 
-# not use @rbe_default. 
+# However, tests covered by this file have a bazerc file that did
```
alex1545: bazerc -> bazelrc

nlopezgi: done

PR #564, opened 2019-06-13
change one more test to use external repo

nlopezgi:


nlopezgi:
blocking https://github.com/bazelbuild/bazel-toolchains/pull/560

PR #575, opened 2019-06-18
Remove use of debian8

nlopezgi:
debian8 is no longer supported with bazel 0.27.0, switch tests that use a custom container to use the bazel cloud builder. This PR is ready to merge as soon as CI passes.


nlopezgi:
this cannot pass until configs are updated, abandoning in favor of https://github.com/bazelbuild/bazel-toolchains/pull/574

PR #583, opened 2019-06-18
fix generation of versions.bzl

nlopezgi:


nlopezgi:
i can't think of a way to test this. the .bzl file produced is valid, so not sure how to make a test that checks that file fail

PR #588, opened 2019-06-19
Add small java sample to all RBE tests

nlopezgi:


tests/rbe_repo/generate_confs.yaml
```diff
@@ -41,6 +40,7 @@ steps:
   - --host_platform=@rbe_autoconf_config_repos//config:platform
   - --platforms=@rbe_autoconf_config_repos//config:platform
   - --remote_instance_name=projects/asci-toolchain/instances/default_instance
+  - //examples/remotebuildexecution/hello_world/cc:say_hello_test
```
xingao267: do we wanna run the java example here as well?

nlopezgi: The tests that are still only building the cc:say_hello_test are ones that use rbe_autoconfig targets in the top level repo, dont want to pollute the top level repo with the java-examples repo. 
I might consider moving some of these tests from using the top level WORKSPACE to use the one in the example repo, but if I do it will be in an upcoming cl.

PR #590, opened 2019-06-20
add test for ubsan

nlopezgi:
Test can only run with bazel 0.23.0 - 0.26.0 (setting it with 0.26.0). Opening up an issue in Bazel repo for advice to update toolchain config for 0.27.0


nlopezgi:
Created https://github.com/bazelbuild/bazel/issues/8687 to followup on making this work with bazel 0.27.0

PR #638, opened 2019-07-17
Add load statement for rules_java.

iirina:


nlopezgi:
Hi Irina,
Please let me know how I can help you with this change.
Some pointers / questions:
- Do not modify any files under configs/ all of these are generated and/or need to work with older versions of Bazel. Any change to make them compatible with Bazel 0.28.x or above would likely break usage of them with older versions of Bazel. Note these targets are only used in CI with the corresponding version of Bazel that they should work with, so should not be a problem if they are broken.
- Make sure your fork is up to date, looks like many of the checks failed to find .yaml files we have added in the last few weeks
- Does it make sense to add the new `rules_java` and `rules_cc` in https://github.com/bazelbuild/bazel-toolchains/blob/master/repositories/repositories.bzl instead of WORKSPACE?
- All the BUILD files you are modifying are now generated. They are generated based on this template: https://github.com/bazelbuild/bazel-toolchains/blob/master/rules/rbe_repo/BUILD.java.tpl I think you need to also make your change there. But just to be sure, is Bazel 0.28.x compatible with using rules_java, even without the --incompatible_load_java_rules_from_bzl flag? If not then we should not change the template before the last version of Bazel that is not compatible with using rules_java is released. This is because our pipeline to generate and release these files is triggered automatically with Bazel releases, so if e.g., 0.28.1 comes out and it does not work with rules_java w/o the --incompatible_load_java_rules_from_bzl, we might break customers of this repo. 

Please ping me if you have questions

nlopezgi:
is this PR still relevant or can we close?

nlopezgi:
Closing, please let me know to reopen if needed.

Issue #649, opened 2019-07-31
rbe_autoconfig with create_java_configs=False failed in verify


lizan:
I recently dropped JDK from our RBE toolchain image, so I tried to generate config without java_home (or use remote_jdk) though it fails in verification:

I got some workaround:
https://github.com/bazelbuild/bazel-toolchains/compare/master...lizan:java_home_check_fix?expand=1 though obviously this is not enough to keep existing configs in versions.bzl.

```
ERROR: /home/lizan/github/lizan/envoy/build/WORKSPACE:17:1: Traceback (most recent call last):
        File "/home/lizan/github/lizan/envoy/build/WORKSPACE", line 17
                envoy_dependency_imports()
        File "/home/lizan/github/lizan/envoy/build/bazel/dependency_imports.bzl", line 12, in envoy_dependency_imports
                rbe_toolchains_config()
        File "/home/lizan/.cache/bazel/_bazel_lizan/935de5c3a9ad4318c27353ce3b82b735/external/envoy/bazel/toolchains/rbe_toolchains_config.bzl", line 74, in rbe_toolchains_config
                _envoy_rbe_toolchain("rbe_ubuntu_clang", _CLANG_ENV, "cla...")
        File "/home/lizan/.cache/bazel/_bazel_lizan/935de5c3a9ad4318c27353ce3b82b735/external/envoy/bazel/toolchains/rbe_toolchains_config.bzl", line 47, in _envoy_rbe_toolchain
                rbe_autoconfig(name = (name + "_gen"), export_con..., <8 more arguments>)
        File "/home/lizan/.cache/bazel/_bazel_lizan/935de5c3a9ad4318c27353ce3b82b735/external/bazel_toolchains/rules/rbe_repo.bzl", line 915, in rbe_autoconfig
                validate_toolchain_config_suite_spec(name, toolchain_config_suite_spec)
        File "/home/lizan/.cache/bazel/_bazel_lizan/935de5c3a9ad4318c27353ce3b82b735/external/bazel_toolchains/rules/rbe_repo/toolchain_config_suite_spec.bzl", line 123, in validate_toolchain_config_suite_spec
                _validate_toolchain_config_suite_autogen_spec(name, toolchain_config_suite_spec)
[...]

PR #650, opened 2019-07-31
enable use of checked in configs with no java home set

nlopezgi:
+ add GCB test that uses `create_java_configs=false`
Fixes https://github.com/bazelbuild/bazel-toolchains/issues/649


lizan:
I believe this still fails when `versions.bzl` is not empty. That's why I added [this line](https://github.com/bazelbuild/bazel-toolchains/compare/master...lizan:java_home_check_fix?expand=1#diff-e943bdb0d4ac905458dba53acfb137e0R1046) in my workaround, it fails with string_list check.

PR #664, opened 2019-08-09
Add rules/exec_properties.bzl

eytankidron:
This bzl files contains functions to create a dictionary of key/values which can be set as platform exec_properties and can be consumed by RBE.


rules/exec_properties.bzl
```diff
@@ -0,0 +1,134 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
nlopezgi: lets move these to a subdirectory under rules
could be 
rules/rbe or something like that

rules/exec_properties.bzl
```diff
@@ -0,0 +1,134 @@
...
+
+def _transform_network(value):
+    return "standard" if value else "off"
+
+
+def create_exec_properties_dict(
+    # The asterisk means that all following arguments must be specified by name at the call site.
+    # This allows future insertions of arguments to this function without breaking callers which might have otherwise depended on argument ordering.
+    *,
+    container_image=None,
```
nlopezgi: put args in alphabetical order?

rules/exec_properties.bzl
```diff
@@ -0,0 +1,134 @@
...
+    pool=None,
+    os_family=None,
+    gce_machine_type=None,
+    jdk_version=None,
+    docker_add_capabilities=None,
```
nlopezgi: run buildifier on these files

rules/exec_properties.bzl
```diff
@@ -0,0 +1,134 @@
...
+    docker_privileged=None,
+    docker_run_as_root=None,
+    docker_runtime=None,
+    docker_sibling_containers=None,
+    docker_ulimits=None,
+    docker_use_urandom=None):
+    """Return a dict with exec_properties that are supported by RBE.
+    
+    Args:
+    container_image (string): The container that RBE will execute in.
```
nlopezgi: you probably do not need to receive the container (or output the container property) as rbe_autoconfig already produces a platform that has this set.
i.e., I expect the use case for the produced dict is to use it in a platform that extends a platform produced by rbe_autoconfig. For reference, this is the template used by rbe_autoconfig: https://github.com/bazelbuild/bazel-toolchains/blob/master/rules/rbe_repo/BUILD.platform.tpl#L38 and this is a sample output platform target: https://github.com/bazelbuild/bazel-toolchains/blob/master/configs/ubuntu16_04_clang/9.0.0/bazel_0.28.1/config/BUILD#L36

rules/exec_properties.bzl
```diff
@@ -0,0 +1,134 @@
...
+    pool (string): If specified, the action will be executed only on this worker pool.
```
nlopezgi: so I'm not sure using "the action will be" is clear here as this macro does not actually run any actions. How about:
"the dict returned will include a property that restricts execution on the given worker pool"
same for all descriptions below

rules/exec_properties.bzl
```diff
@@ -0,0 +1,134 @@
...
+    os_family (string): If specified, the action will be executed only on workers of this OS family.
+    gce_machine_type (string): If specified, the action will be executed only on workers of this machine type.
+    jdk_version (string): If specified, the action will be executed only on workers that have this JDK version.
+    docker_add_capabilities (string): If specified, RBE will add these (comma separated) capabilities to the capabilities it uses for running docker.
```
nlopezgi: If specified, the dict produced will include these (comma separated) capabilities to the capabilities selected for running docker.

rules/exec_properties.bzl
eytankidron: Now that I think about it, I think it might be better to just remove these comments explaining what each argument is and instead just point to https://cloud.google.com/remote-build-execution/docs/remote-execution-environment#remote_execution_properties since each of these arguments maps to an argument that is documented there.

rules/exec_properties.bzl
eytankidron: container_image, just like any of these parameters, is optional.
I agree that it might not be useful when using it in rbe_autoconfig, but I envision this macro might also be useful for creating a platform without going through rbe_autoconfig.

Does that make sense?

rules/exec_properties.bzl
eytankidron: I see that there is a directory rules/rbe_repo/ does it make sense to put it there or would you rather I create a new directory rules/rbe/ ?

rules/exec_properties.bzl
eytankidron: Done.

rules/exec_properties.bzl
eytankidron: Done.

rules/exec_properties.bzl
eytankidron: I removed these comments.

rules/exec_properties.bzl
eytankidron: I'm removing this entire section and pointing instead to https://cloud.google.com/remote-build-execution/docs/remote-execution-environment#remote_execution_properties
 which is better documentation anyway.

rules/exec_properties.bzl
eytankidron: Following our offline conversation, I'm putting this in rules/experimental/rbe/ for now.

rules/exec_properties.bzl
eytankidron: Following our offline conversation, I'm going to leave container_image in for now.
We should follow this up at some point once we have a better understanding of the use cases. We should also document these use cases somewhere. If there is truly no use case for using container_image, we can remove it later. Since this code is now under experimental, we can do this.

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -0,0 +1,163 @@
...
+    Args:
+      **kwargs: Arguments specifying what keys are populated in the returned dict.
+          Note that the name of the key in kwargs is not the same as the name of the key in the returned dict.
+          For more information about what each parameter is see https://cloud.google.com/remote-build-execution/docs/remote-execution-environment#remote_execution_properties.
+          If this link is broken for you, you may not to be whitelisted for RBE. See https://groups.google.com/forum/#!forum/rbe-alpha-customers.
+
+    Returns:
+      A dict that can be used as, for example, the exec_properties parameter of platform.
+    """
+    params = {
```
nlopezgi: you can probably declare this statically outside of the def. But that can be fine in an upcoming PR to unblock you.

nlopezgi:
i'm rerunning the build that failed due to timeout, seems to be flaky as of lately

rules/experimental/rbe/exec_properties.bzl
eytankidron: Done.

buchgr:
@eytankidron what's the future grand plan for this? Is there a design doc I could look at?

I am wondering if we should chat about having a more abstract API for configuring common use cases and possibly putting the interface in a more neutral place like bazelbuild/platforms with RBE specific implementation being in bazel-toolchains.

eytankidron:
> What's the future grand plan for this? Is there a design doc I could look at?

There is no design doc yet, although I might write something up at some point.
The main rationale for this macro is to provide a way for users to avoid typos in property names, which are really hard to detect because RBE generally silently ignores any properties it does not recognize.

> I am wondering if we should chat about having a more abstract API for configuring common use cases and possibly putting the interface in a more neutral place like bazelbuild/platforms with RBE specific implementation being in bazel-toolchains.

I'm not sure what the more abstract part would be that we would want to put somewhere else. But I'm happy to chat about this. I'll set something up.
The current tentative plan is to keep this in bazelbuild/bazel-toolchains, but obviously move it out of an experimental directory once this is more hardened, tested and documented.

PR #669, opened 2019-08-16
Add create_exec_properties_dict tests

eytankidron:
Create some tests that run on RBE with platforms with various exec_properties that are populated using create_exec_properties_dict()

Currently, I added only tests for docker_privileged=True/False and for docker_network_enabled=True/False. But this PR provides a skeleton that demonstrates how other properties should be tested.

In order to run these tests, bazel needs to run with
--extra_execution_platforms=@rbe_default//config:platform,//tests/rules/experimental/rbe:platform_privileged_on,//tests/rules/experimental/rbe:platform_network_enabled

I have not yet piped these test into any pipeline.
One option is to pipe these tests as presubmit tests of bazel-toolchains. This cannot yet be done because these tests can only run on bazel 0.29 and up and that container does not yet exist. The reason why it is not compatible with older bazel versions is because platform does not have an exec_properties in those versions.

However, what I really want to do is to pipe these tests as bazel buildkite downstream tests (see https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#configuring-a-pipeline).
In order to do that, .bazelci/presubmit.yml needs to first be converted to a newer format. I plan to do this in a separate pull request.

### DIFF
--- a/.bazelversion
+++ b/.bazelversion
@@ -1,1 +1,0 @@
-last_downstream_green
--- a/.bazelversion
+++ b/.bazelversion
@@ -1,0 +1,1 @@
+last_downstream_green
--- a/tests/rules/experimental/rbe/BUILD
+++ b/tests/rules/experimental/rbe/BUILD
@@ -1,5 +1,21 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 load("//rules/experimental/rbe:exec_properties.bzl", "create_exec_properties_dict")
 load("//configs/ubuntu16_04_clang:versions.bzl", "LATEST")
+
+CONTAINER_IMAGE = "docker://l.gcr.io/google/rbe-ubuntu16-04@%s" % LATEST
 
 # Test the docker_privileged property
 constraint_setting(name = "constraint_privileged")
@@ -13,7 +29,7 @@ platform(
     name = "platform_privileged_on",
     constraint_values = [":constraint_privileged_on"],
     exec_properties = create_exec_properties_dict(
-        container_image = "docker://marketplace.gcr.io/google/rbe-ubuntu16-04@%s" % LATEST,
+        container_image = CONTAINER_IMAGE,
         docker_privileged = True,
     ),
 )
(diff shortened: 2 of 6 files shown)


tests/rules/experimental/rbe/BUILD
```diff
@@ -0,0 +1,66 @@
+load("//rules/experimental/rbe:exec_properties.bzl", "create_exec_properties_dict")
```
nlopezgi: add license header

tests/rules/experimental/rbe/BUILD
```diff
@@ -0,0 +1,66 @@
...
+constraint_value(
+    name = "constraint_privileged_on",
+    constraint_setting = ":constraint_privileged",
+)
+
+platform(
+    name = "platform_privileged_on",
+    constraint_values = [":constraint_privileged_on"],
+    exec_properties = create_exec_properties_dict(
+        container_image = "docker://marketplace.gcr.io/google/rbe-ubuntu16-04@%s" % LATEST,
```
nlopezgi: this should likely point to l.gcr.io because in cases in which rbe_autoconfig needs to pull the container it will fail to do so in some CI pipelines because marketplace.gcr.io requires auth, whereas l.gcr.io does not.

tests/rules/experimental/rbe/network_disabled.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
nlopezgi: add License header

tests/rules/experimental/rbe/network_disabled.sh
```diff
@@ -0,0 +1,9 @@
...
+
```
nlopezgi: add short doc with purpose of this script

tests/rules/experimental/rbe/network_enabled.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
nlopezgi: same as above, license + short doc, same for the two other scripts below

tests/rules/experimental/rbe/BUILD
eytankidron: Done.

tests/rules/experimental/rbe/BUILD
eytankidron: Done.

tests/rules/experimental/rbe/network_disabled.sh
eytankidron: Done.

tests/rules/experimental/rbe/network_disabled.sh
eytankidron: Done.

tests/rules/experimental/rbe/network_enabled.sh
eytankidron: Done.

.bazelversion
```diff
@@ -0,0 +1 @@
+last_downstream_green
```
nlopezgi: will this mean all our tests will run with the last_downstream_green? Will this mean we get the latest features (and breaking features)? Not sure we want to commit to that at the top level directory. if you want this for your tests you could add a sub project (e.g., https://github.com/bazelbuild/bazel-toolchains/tree/master/tests/rbe_repo/examples/bazel_toolchains_client and add its top level dir to https://github.com/bazelbuild/bazel-toolchains/blob/master/.bazelignore.

nlopezgi: (sorry approved by mistake, lets figure out if we can remove this file from this PR or move it to a sub project)

eytankidron: My bad. I did not mean to add this to the PR.
I will remove it.

PR #670, opened 2019-08-16
Modify presubmit.yml to use the new format.

eytankidron:
I'm basing this change on the instructions in https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#legacy-format

This change should not change anything. As in, the exact same tests should run on any upstream change.

Not at all sure how to test this :/

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -8,11 +8,11 @@ buildifier:
   # TODO (suvanjan): Re-enable once issues and https://github.com/bazelbuild/buildtools/issues/575
   # are fixed.
   warnings: "attr-cfg,attr-license,attr-non-empty,attr-output-default,attr-single-file,confusing-name,constant-glob,ctx-actions,ctx-args,depset-iteration,depset-union,dict-concatenation,duplicated-name,filetype,function-docstring,git-repository,http-archive,integer-division,load,load-on-top,module-docstring,name-conventions,native-build,native-package,no-effect,output-group,package-name,package-on-top,positional-args,redefined-variable,repository-name,same-origin-load,string-iteration,unreachable,unused-variable"
-platforms:
-  rbe_ubuntu1604:
+tasks:
+  rbe_autoconfig:
+    platform: ubuntu1604
     test_targets:
     - "--"
-    # rbe_autoconfig tests
     - "//tests/rbe_repo/..."
     # tests that require output_base / config_repos are timing out.
     # TODO(nlopezgi): debug, fix and enable these tests.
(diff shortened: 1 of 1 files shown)


eytankidron:
@philwo, could you please take a look at this PR and let me know if it looks right.
Also, I would love to know: how do I debug this change and verify that it actually works?

.bazelci/presubmit.yml
```diff
@@ -25,7 +25,19 @@ platforms:
     - "-//tests/rbe_repo:rbe_autoconf_config_repos_sh_toolchain_test"
     - "-//tests/rbe_repo:rbe_autoconf_config_repos_no_cc_config_run_in_container_test"
     - "-//tests/rbe_repo:rbe_autoconf_config_repos_no_cc_config_sh_toolchain_test"
-    # docker_autoconfig tests
+    test_flags:
+    - "--action_env=RBE_AUTOCONF_ROOT=/workdir"
+    - "--action_env=PATH"
+    - "--test_output=errors"
+    - "--verbose_failures"
+    - "--extra_toolchains=@bazel_toolchains//configs/ubuntu16_04_clang/latest:toolchain_docker"
```
nlopezgi: This is wrong. Targets under //configs/ubuntu16_04_clang/latest are not in maintenance and should never be used.
You should always use rbe_autoconfig targets to access configs in bazel_toolchains repo. 
Documentation for (buildkite CI with RBE)[https://docs.bazel.build/versions/master/remote-execution-ci.html#setting-up-the-bazel-ci-for-testing] indicates you should be using an rbe autoconfig rule:
```
# Creates toolchain configuration for remote execution with BuildKite CI
# for rbe_ubuntu1604.
rbe_autoconfig(name = "buildkite_config")
```
The expectation is that BuildKite CI is already configured with the right flags to use RBE with this @buildkite_config target.
If this has changed, then we need an update on the (docs)[https://docs.bazel.build/versions/master/remote-execution-ci.html#setting-up-the-bazel-ci-for-testing] and should still be using @buildkite_config targets here.

.bazelci/presubmit.yml
```diff
@@ -25,7 +25,19 @@ platforms:
...
+    - "--extra_execution_platforms=@bazel_toolchains//configs/ubuntu16_04_clang/latest:platform_docker"
+    - "--host_platform=@bazel_toolchains//configs/ubuntu16_04_clang/latest:platform_docker"
+    - "--platforms=@bazel_toolchains//configs/ubuntu16_04_clang/latest:platform_docker"
+  docker_autoconfig:
```
nlopezgi: Not sure why you decided to break off the docker_autoconfig to run on a non rbe platform? unless you have a reason, keep them together with the same platform as above.

.bazelci/presubmit.yml
eytankidron: It was not my intention to change this behavior in this PR. I just split the existing task into two.
However, given that these flags are deprecated, it's better that I don't.
I'm going to abandon this PR in favor of https://github.com/bazelbuild/bazel-toolchains/pull/671.

.bazelci/presubmit.yml
eytankidron: No particular reason. I basically wanted this PR to demonstrate that the new format, using task instead of platform, can support multiple of these tasks.
I will abandon this PR and instead go ahead with https://github.com/bazelbuild/bazel-toolchains/pull/671 which adds a separate new task .

nlopezgi:
Closing in favor of https://github.com/bazelbuild/bazel-toolchains/pull/671

PR #671, opened 2019-08-19
Adding a rbe_exec_properties task to presubmit.yml

eytankidron:


.bazelci/presubmit.yml
```diff
@@ -45,3 +45,14 @@ platforms:
     - "--extra_execution_platforms=@bazel_toolchains//configs/ubuntu16_04_clang/latest:platform_docker"
     - "--host_platform=@bazel_toolchains//configs/ubuntu16_04_clang/latest:platform_docker"
     - "--platforms=@bazel_toolchains//configs/ubuntu16_04_clang/latest:platform_docker"
+  rbe_exec_properties:
```
nlopezgi: add short doc comment above explaining what is being tested here differently from above.

eytankidron: Done.

nlopezgi:
looks like buildkite build failed (maybe it didn't like the location of your comment?)

eytankidron:
> looks like buildkite build failed
I changed the platform of the new task to be rbe_ubuntu1604 (now both tasks have that platform).
Let's see if this fixes the problem.

eytankidron:
Looks like buildkite is running these new test on bazel 0.28.1 (latest) and they are, unsurprisingly, failing because exec_properties was added to platform only at 0.29 which isn't released yet.

@philwo, is there a way to annotate the task in such a way that it will run on, say, 0.29.0+ (which is my way of saying 0.29.0 and up)? Or, even better, 0.29.0rc6+ ?

PR #711, opened 2019-09-20
Add rbe_exec_properties and custom_exec_properties

eytankidron:
These are starlark macros that create repo rules.
rbe_exec_properties creates a repo rule that contains standard RBE execution property sets.
custom_exec_properties creates a repo rule that contains custom execution property sets.

The PR also includes documentation reagrding how these repo rules should be used and why.


eytankidron:
@agoulti @buchgr could you please review this change request?
This is based on the ideas we discussed a few weeks ago.

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -13,6 +13,92 @@
...
+
+  load("@exec_properties//:exec_propert_sets.bzl", "NETWORK_ON")
+  ...
+  exec_properties = NETWORK_ON
+
+The reason not to directly set exec_properties = {...} in a target is that then it might be hard to
+depend on such a target from another repo, if, say, that other repo wants to use remote execution
+but not RBE.
+
+Scenario 2 - non-RBE remote execution:
```
nlopezgi: How would this apply for extending a repo with local execution only, extending from a repo that uses @exec_properties? Is it relevant to include that scenario here?

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -13,6 +13,92 @@
...
+
+And then in the BUILD file:
+  load("@my_bespoke_exec_properties//:exec_propert_sets.bzl", "HIGH_MEM_MACHINE")
+  target(
+      ...
+      exec_properties = HIGH_MEM_MACHINE,
+  )
+
+A depending repo can then either define HIGH_MEM_MACHINE on @my_bespoke_exec_properties to be
+{"gceMachineType" : "n1-highmem-8"}, or it can define it to be anything else.
```
nlopezgi: Can you add here the example where its set to "nothing so that it just runs locally fine"? (and just doing this might also address well enough my first comment above)

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -156,3 +242,105 @@ def merge_dicts(*dict_args):
     for dictionary in dict_args:
         result.update(dictionary)
     return result
+
+def _exec_property_sets_repository_impl(repository_ctx):
+    repository_ctx.file(
+        "BUILD",
+        content = "",
```
nlopezgi: Would it be relevant to have this BUILD file have a bzl_library target?

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -156,3 +242,105 @@ def merge_dicts(*dict_args):
...
+        content = repository_ctx.attr.exec_property_sets_content,
+        executable = False,
+    )
+
+# _exec_property_sets_repository is a repository rule that creates a repo with the specified exec_property_sets.
+_exec_property_sets_repository = repository_rule(
+    implementation = _exec_property_sets_repository_impl,
+    local = True,
+    attrs = {
+        "exec_property_sets_content": attr.string(mandatory = True),
```
nlopezgi: add a doc attribute to exec_property_sets_content and explain what its for here

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -156,3 +242,105 @@ def merge_dicts(*dict_args):
...
+    exec_property_sets_content = ""
+    for key, value in dicts.items():
+        exec_property_sets_content += "%s = %s\n" % (key, value)
+
+    _exec_property_sets_repository(
+        name = name,
+        exec_property_sets_content = exec_property_sets_content,
+    )
+
+standard_property_sets = {
```
nlopezgi: capitalize this const name?

tests/rules/experimental/rbe/BUILD
```diff
@@ -12,9 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("//rules/experimental/rbe:exec_properties.bzl", "create_exec_properties_dict")
+load("@exec_properties//:exec_propert_sets.bzl", "DOCKER_PRIVILEGED", "NETWORK_ON")
```
nlopezgi: we're loosing test coverage of create_exec_properties_dict. Is it necessary to keep it? Is now create_exec_properties_dict an implementation detail? If so, can we document it as so?

rules/experimental/rbe/exec_properties.bzl
eytankidron: I actually thought about it, but it seems that it works just fine without defining such a target.
What would be the reason to define this as a target?

tests/rules/experimental/rbe/BUILD
eytankidron: It's not exactly an implementation detail.
We are still calling it, but now we are doing so from the WORKSPACE file.
For example, it was used in the creation of exec_properties_with_override and network_on_exec_properties.

rules/experimental/rbe/exec_properties.bzl
eytankidron: Local execution ignores the exec_properties field completely, so it doesn't actually matter how it is populated.
I will add a scenario that states that explicitly.

rules/experimental/rbe/exec_properties.bzl
eytankidron: See my previous comment.

rules/experimental/rbe/exec_properties.bzl
eytankidron: done

rules/experimental/rbe/exec_properties.bzl
eytankidron: done

WORKSPACE
```diff
@@ -578,3 +578,27 @@ rbe_autoconfig(
...
+
+# Define several exec property repo rules to be used in testing
+load("//rules/experimental/rbe:exec_properties.bzl", "create_exec_properties_dict", "custom_exec_properties", "rbe_exec_properties")
+
+# A standrad RBE execution property set repo rule
+rbe_exec_properties(
+    name = "exec_properties",
+)
+
+# A standrad RBE execution property set repo rule whose NETWORK_ON has been overridden by "network off".
```
agoulti: /standrad/standard/

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -13,6 +13,97 @@
...
+      name = "exec_properties",
+  )
+
+And the targets in the BUILD files will be able to depend on targets from other repos that were
+written with RBE in mind.
+
+Scenario 4 - rbe_exec_properties with override:
+
+Let's now assume that a particular repo, running with a particular RBE setups, wants to enforce
+for the sake of hermeticity, that no build or test will ever have network access. This would be
```
agoulti: I feel that the word "enforce" is a little too strong. This does not guarantee that there won't be something that turns it on. The rules might e.g., still put this into a dictionary directly.

Maybe "wants to run everything without network access"

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -156,3 +247,108 @@ def merge_dicts(*dict_args):
...
+    "NOT_DOCKER_PRIVILEGED": create_exec_properties_dict(docker_privileged = False),
+    "DOCKER_RUN_AS_ROOT": create_exec_properties_dict(docker_run_as_root = True),
+    "NOT_DOCKER_RUN_AS_ROOT": create_exec_properties_dict(docker_run_as_root = False),
+    "DOCKER_SIBLINGS_CONTAINERS": create_exec_properties_dict(docker_sibling_containers = True),
+    "NOT_DOCKER_SIBLINGS_CONTAINERS": create_exec_properties_dict(docker_sibling_containers = False),
+    "DOCKER_USE_URANDOM": create_exec_properties_dict(docker_use_urandom = True),
+    "NOT_DOCKER_USE_URANDOM": create_exec_properties_dict(docker_use_urandom = False),
+    "LINUX": create_exec_properties_dict(os_family = "Linux"),
+    "WINDOWS": create_exec_properties_dict(os_family = "Windows"),
+    "NVIDIA": create_exec_properties_dict(docker_runtime = "nvidia"),
```
agoulti: Should we remove this from here?
I'll follow up with a doc, but basically: I don't think a user would ever want to use this standalone, currently they need this together with some way to identify the pool.
Let's talk more

WORKSPACE
eytankidron: Done.

rules/experimental/rbe/exec_properties.bzl
eytankidron: Done.

rules/experimental/rbe/exec_properties.bzl
eytankidron: Done.

rules/experimental/rbe/exec_properties.bzl
nlopezgi: These targets need to be defined if a project depends on this rule from another rule and wants to, e.g., use skylib to generate docs.

tests/rules/experimental/rbe/BUILD
nlopezgi: If its still being used, should we still have some tests that exercise it independently of the new rule? It seems its only now being tested indirectly via the repo rules.

WORKSPACE
```diff
@@ -578,3 +578,27 @@ rbe_autoconfig(
 # Needed for testing purposes. Creates a file that exposes
 # the value of RBE_AUTOCONF_ROOT
 rbe_autoconfig_root(name = "rbe_autoconfig_root")
+
+# Define several exec property repo rules to be used in testing
```
nlopezgi: nit: add '.' here and below

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -13,6 +13,96 @@
 # limitations under the License.
 
 """This file contains macros to create and manipulate dictionaries of properties to be used as execution properties for RBE.
+
+It also contains macros that create repository rules for standard and custom sets of execution
```
nlopezgi: Should we have this statement first and the "macros to create and manipulate dictionaries of properties" second?

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -13,6 +13,96 @@
...
+Then, in some BUILD file, you can reference this execution property constant as follows:
+
+  load("@exec_properties//:constants.bzl", "NETWORK_ON")
+  ...
+  exec_properties = NETWORK_ON
+
+The reason not to directly set exec_properties = {...} in a target is that then it might be hard to
+depend on such a target from another repo, if, say, that other repo wants to use remote execution
+but not RBE.
+
```
nlopezgi: Should we have a pointer here to the SoT for which are the "standard RBE execution property constants"

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -13,6 +13,96 @@
...
+Scenario 2 - local execution
+
+If bazel is set up so that the targets are executed locally, then the contents of exec_properties
```
nlopezgi: Capitalize Bazel

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -13,6 +13,96 @@
...
+Let's assume that the non-RBE remote execution endpoint provides a macro similar to
+rbe_exec_properties (say other_re_exec_properties), which populates the same constants (e.g.
+NETWORK_ON) with possibly different dict values.
+In this case, the WORKSPACE would look like this:
+  other_re_exec_properties(
+      name = "exec_properties",
+  )
+
+And the targets in the BUILD files will be able to depend on targets from other repos that were
+written with RBE in mind.
```
nlopezgi: written with RBE in mind as the name of the targets in the WORKSPACE (exec_properties in this case) is the same...
Or something to make it super explicit that this works if proper naming conventions are followed.

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -13,6 +13,96 @@
...
+
+Scenario 4 - rbe_exec_properties with override:
+
+Let's now assume that a particular repo, running with a particular RBE setups, wants to run
```
nlopezgi: setup

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -13,6 +13,96 @@
...
+
+In the WORKSPACE file, call
+  rbe_exec_properties(
+      name = "exec_properties",
+      override = {
+          "NETWORK_ON": create_exec_properties_dict(docker_network = "off"),
+      },
+  )
+
+This would override the meaning of NETWORK_ON for this workspace only.
```
nlopezgi: again, add note about naming of target matching

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -13,6 +13,96 @@
...
+
+Scenario 5 - custom execution properties
+
+In this scenario, let's assume that a target is best run remotely on a high memory GCE machine.
+The RBE setup associated with the workspace where the target is defined has workers of type
+"n1-highmem-8".
+Setting exec_properties = {"gceMachineType" : "n1-highmem-8"} is problematic because it does not
+lend itself to another repo depending on this target if, for example, the other repo does not use
+RBE. Or it might use RBE but have a different high memory GCE machine such as "n1-highmem-16".
```
nlopezgi: s/does not use RBE/ uses a remote execution endpoint other than RBE
(make it explicit this applies for remote exec only)

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -13,6 +13,96 @@
...
+Unlike the case of NETWORK_ON, rbe_exec_properties does not provide a standard HIGH_MEM_MACHINE
+execution property set (although it might do so in the future).
+
+The recommended way to do this is as follows:
```
nlopezgi: what is 'this' exactly referring to in this sentence?

tests/rules/experimental/rbe/BUILD
eytankidron: I'm inclined to say no.
Calling create_exec_properties_dict() from a BUILD file, while not impossible, is an anti-pattern.
I'm actually interested in putting in place some kind of linter check (maybe in buildifier?) that would prevent that anti-pattern, or maybe just give out a warning.

nlopezgi: I'm not referring to whether customers should call this from a BUILD file or not. I'm referring to how we are testing the rule from within this repo. Prior to this PR, we had some form of unit tests that directly checked the create_exec_properties_dict rule, with this PR we do not, we have tests that indirectly depend on create_exec_properties_dict. Lets re add them, even if users should not ever call create_exec_properties_dict from a BUILD file

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -13,6 +13,96 @@
...
+
+And then in the BUILD file:
+  load("@my_bespoke_exec_properties//:constants.bzl", "HIGH_MEM_MACHINE")
+  target(
+      ...
+      exec_properties = HIGH_MEM_MACHINE,
+  )
+
+A depending repo can then either define HIGH_MEM_MACHINE on @my_bespoke_exec_properties to be
+{"gceMachineType" : "n1-highmem-8"}, or it can define it to be anything else.
```
nlopezgi: what is a valid example of "anything else" here?

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -156,3 +246,107 @@ def merge_dicts(*dict_args):
...
+    constants_bzl_content = ""
+    for key, value in dicts.items():
+        constants_bzl_content += "%s = %s\n" % (key, value)
+
+    _exec_property_sets_repository(
+        name = name,
+        constants_bzl_content = constants_bzl_content,
+    )
+
+STANDARD_PROPERTY_SETS = {
```
nlopezgi: Add a short doc comment mentioning this is the SoT for the properties used by RBE

nlopezgi: Is there a pointer to somewhere that explains each of these? add pointer here if it exists, document properties here if it does not.

rules/experimental/rbe/exec_properties.bzl
```diff
@@ -156,3 +246,107 @@ def merge_dicts(*dict_args):
...
+                fail("In repo rule %s, execution property set %s, key %s, the value %s must be a string" % (name, key, k, v))
+
+def custom_exec_properties(name, dicts):
+    """ Creates a repository containing execution property dicts.
+
+    Use this macro in your WORKSPACE.
+
+    Args:
+      name: Name of the repo rule.
+      dicts: The execution property set constants.
```
nlopezgi: explain in more detail the expected data structure

tests/rules/experimental/rbe/BUILD
eytankidron: But then if and when we add this linter check to buildifier, this particular build file will fail the lint test.

nlopezgi: We can always turn off that check for this particular BUILD file

rules/experimental/rbe/exec_properties.bzl
eytankidron: done

tests/rules/experimental/rbe/BUILD
eytankidron: Following our offline discussion, I agree that we *could* turn off the check for this particular BUILD file.
The question we need to answer, I think, is would testing create_exec_properties_dict() directly from the BUILD file (despite being an anti-pattern) add extra test coverage that we don't otherwise have.
If the answer is yes, then I am happy to make a linter exception here to improve code coverage. But I don't think it does add more coverage.

WORKSPACE
eytankidron: done

rules/experimental/rbe/exec_properties.bzl
eytankidron: done

rules/experimental/rbe/exec_properties.bzl
eytankidron: done

rules/experimental/rbe/exec_properties.bzl
eytankidron: done

rules/experimental/rbe/exec_properties.bzl
eytankidron: done

rules/experimental/rbe/exec_properties.bzl
eytankidron: done

rules/experimental/rbe/exec_properties.bzl
eytankidron: done
(I added a sentence, which I *think* addresses your comment.)

rules/experimental/rbe/exec_properties.bzl
eytankidron: done

rules/experimental/rbe/exec_properties.bzl
eytankidron: done

rules/experimental/rbe/exec_properties.bzl
eytankidron: done

rules/experimental/rbe/exec_properties.bzl
eytankidron: Done.
Added a comment saying that this is the SoT for the list of standard constants.
As for the docs of the specific properties, I added a link to https://cloud.google.com/remote-build-execution/docs/remote-execution-environment#remote_execution_properties.
This link currently contains only the docker related properties, but we should update our RBE docs to contain all the properties.
We made a decision *not* to duplicate the documentation of each property for the sake of easier maintenance. We also decided to have the property names as received by create_exec_properties_dict() match exactly the RBE names with the only difference being that they are in snake_case.

rules/experimental/rbe/exec_properties.bzl
eytankidron: done

PR #712, opened 2019-09-25
add warning if current running bazel is ahead of bazel-toolchains repo

nlopezgi:
This warning should enable users to know when its unlikely to get toolchain configs that are checked in after a bazel upgrade.

### DIFF
--- a/rules/rbe_repo/version_check.bzl
+++ b/rules/rbe_repo/version_check.bzl
@@ -60,10 +60,12 @@ def _check_bazel_version(bazel_version_fallback):
                "was not defined explicitly in rbe_autoconfig target. " +
                "Falling back to '%s'") % bazel_version_fallback)
         return bazel_version_fallback
+
     # If running a release that is not an RC, print a warning if the
     # fallback (latest known in this repo) is old
-    if (native.bazel_version.find("rc")==-1 and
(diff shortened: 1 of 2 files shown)


rules/rbe_repo/version_check.bzl
```diff
@@ -60,4 +60,7 @@ def _check_bazel_version(bazel_version_fallback):
                "was not defined explicitly in rbe_autoconfig target. " +
                "Falling back to '%s'") % bazel_version_fallback)
         return bazel_version_fallback
+    if native.bazel_version > bazel_version_fallback:
+        print("\nCurrent running Bazel is ahead of bazel-toolchains repo. " +
+              "Please update your pin to bazel-toolchains repo.")
```
agoulti: Thanks for doing this! This should help with some of the confusion we've seen.

Maybe add "in your WORKSPACE file" or something similar?
I know it obviously follows from "a pin to a repo", but for those not intimately familiar with Bazel it's not always top of mind. And/or add a link to your Releases page that has the snippet to paste in the WORKSPACE?

nlopezgi: done added the "in your WORKSPACE" will not add any links to release page here to keep the message as short and as unobtrusive as possible (also don't want to make that URL part of our API surface).

PR #715, opened 2019-09-30
Add a bazel 1.0.0 bazelrc file

eytankidron:
This should be submitted only once 1.0.0 is released.

Note that this file is identical to 0.29.0 except for the fact that I added --incompatible_strict_action_env=true


nlopezgi:
closing - rc file is included in https://github.com/bazelbuild/bazel-toolchains/pull/736

Issue #356, opened 2019-03-09
rbe_autoconfig allow extra remote_execution_properties


ebracho:
For our rbe platform we need to configure more than just the container-image (e.g. `dockerNetwork`, `dockerAddCapabilities`, `dockerSiblingContainers`). Right now we have to edit them in after generating the autoconfig. It would be great if we could pass them through the `rbe_autoconfig` rule itself.

nlopezgi:
thanks for opening this issue. we are planning on spring this on the future, for now, the recommended solution is to declare a platform that had as parent the generated platform and which adds the additional properties. example: https://github.com/bazelbuild/rules_docker/blob/master/toolchains/BUILD

ebracho:
That should work perfectly, thanks!

buchgr:
I'd also like to have this!

eytankidron:
As of bazel 1.0 there is a now a way to do this.

Both platforms and rbe_autoconfig now have a field named exec_properties.

In a platform, exec_properties is replacing the deprecated remote_execution_properties. Unlike remote_execution_properties, exec_properties receives a string->string dict and not a protobuf-serialized-as-string.

rbe_autoconfig now also has a new field named use_legacy_platform_definition. For backwards compatibility it is set by default to True, but if you want the underlying platform to use the exec_properties field, rather than remote_execution_properties, you should set it to False. If you want to populate exec_properties directly in rbe_autoconfig, you must set use_legacy_platform_definition to False.

exec_properties receives a string->string dict but do not populate this dict manually. Instead, please use create_exec_properties_dict (see https://github.com/bazelbuild/bazel-toolchains/tree/master/rules/exec_properties). This is a bazel macro that creates this dict for you and makes the proper type checks (and some basic value checks) to ensure that RBE can consume these remote execution properties.

In a related note, as of bazel 1.0 individual targets now also have an exec_properties attribute, which are added to (or override) the exec_properties of the platform for the execution of that particular target. For more on this see https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes. This may remove the need for creating inheriting platforms in many cases. 

The properties that you mentioned (dockerNetwork, dockerAddCapabilities, dockerSiblingContainers) may be set in rbe_autoconfig if you want them for all targets, or you may configure them per target if they are needed only for very specific tests. No need to use toolchain constraints for this to work.

PR #720, opened 2019-10-02
Add an exec_properties attr to rbe_autoconfig

eytankidron:
This attr is only supported when use_legacy_platform_definition is False.
It creates the underlying platform with the given execution properties.
Note that the container image cannot be set via exec_properties. There are other mechanisms for setting the container image and I would rather not cause confusion by adding another way (although we could allow this as an override mechanism in the future if we find it useful).

I modified one of the integration tests in tests/rules/experimental/rbe/BUILD to use a platform that was created using rbe_autoconfig with exec_properties.

I did not yet modify the rbe_autoconfig documentation in rules/rbe_repo.bzl. I'd like to wait with that until we lock down where rbe_exec_properties lives (e.g. not in experimental).

This is a solution to https://github.com/bazelbuild/bazel-toolchains/issues/356


WORKSPACE
```diff
@@ -602,3 +602,12 @@ custom_exec_properties(
         "BESPOKE_NETWORK_ON": create_exec_properties_dict(docker_network = "standard"),
     },
 )
+
+load("@exec_properties//:constants.bzl", "DOCKER_PRIVILEGED")
+
+# Same as @rbe_default_exec_properties execept that the exec properties also include DOCKER_PRIVILEGED.
```
nlopezgi: I was noticing all the examples include turning on a single exec property, how will this work if I need DOCKER_PRIVILEGED and NETWORK_ON?

rules/rbe_repo.bzl
```diff
@@ -696,6 +699,15 @@ _rbe_autoconfig = repository_rule(
                    "example, [\"@bazel_tools//platforms:linux\"]. Default " +
                    " is set to values for rbe-ubuntu16-04 container."),
         ),
+        "exec_properties": attr.string_dict(
+            doc = (
+                "Optional. The execution properties to be used when creating the " +
+                "underlying platform. When providing this attribute, " +
+                "use_legacy_platform_definition must be set to False. Note that " +
+                "the container image property must not be specified via this " +
```
nlopezgi: container_image.
(will this fail if its specified?)

rules/rbe_repo.bzl
```diff
@@ -774,7 +786,8 @@ _rbe_autoconfig = repository_rule(
                 "is important is because a platform that inherits from this " +
                 "platform and wishes to add execution properties must use the " +
                 "same field remote_execution_properties/exec_properties that " +
-                "the parent platform uses."
+                "the parent platform uses. This attribute must be set to False if the " +
```
nlopezgi: Do we currently fail if it's not?

rules/rbe_repo.bzl
```diff
@@ -872,6 +886,10 @@ def rbe_autoconfig(
       exec_compatible_with: Optional. List of constraints to add to the produced
           toolchain/platform targets (e.g., ["@bazel_tools//platforms:linux"] in the
           exec_compatible_with/constraint_values attrs, respectively.
+      exec_properties: Optional. A string->string dict containing execution properties
```
nlopezgi: wrap at the same length as all other lines in this file

rules/rbe_repo/build_gen.bzl
```diff
@@ -94,27 +95,31 @@ def create_export_platform(ctx, image_name, name, toolchain_config_spec_name, us
       name: name of rbe_autoconfig repo rule.
       toolchain_config_spec_name: name of the toolchain config spec
       use_legacy_platform_definition: Whether to create a platform with remote_execution_properties (legacy) or with exec_properties.
+      exec_properties: A string->string dict containing execution properties to be used when creating the platform.
```
nlopezgi: wrap lines

rules/rbe_repo/build_gen.bzl
```diff
@@ -85,7 +86,7 @@ def create_java_runtime(ctx, java_home):
         False,
     )
 
-def create_export_platform(ctx, image_name, name, toolchain_config_spec_name, use_legacy_platform_definition):
+def create_export_platform(ctx, image_name, name, toolchain_config_spec_name, use_legacy_platform_definition, exec_properties):
```
nlopezgi: keep args in alphabetical order

rules/rbe_repo/build_gen.bzl
```diff
@@ -94,27 +95,31 @@ def create_export_platform(ctx, image_name, name, toolchain_config_spec_name, us
...
     cc_toolchain_target = "//" + ctx.attr.toolchain_config_suite_spec["output_base"]
     if toolchain_config_spec_name:
         cc_toolchain_target += "/" + toolchain_config_spec_name
     cc_toolchain_target += "/bazel_" + ctx.attr.bazel_version
     cc_toolchain_target += "/cc" + _CC_TOOLCHAIN
-    _create_platform(ctx, image_name, name, cc_toolchain_target, use_legacy_platform_definition)
+    _create_platform(ctx, image_name, name, cc_toolchain_target, use_legacy_platform_definition, exec_properties)
 
-def create_external_repo_platform(ctx, image_name, name, use_legacy_platform_definition):
+def create_external_repo_platform(ctx, image_name, name, use_legacy_platform_definition, exec_properties):
```
nlopezgi: alphabetical order for args

rules/rbe_repo/build_gen.bzl
```diff
@@ -132,24 +139,38 @@ def create_alias_platform(ctx, toolchain_config_spec_name, image_name, name, use
         target = _CC_TOOLCHAIN,
         toolchain_config_repo = ctx.attr.toolchain_config_suite_spec["repo_name"],
     ))
-    _create_platform(ctx, image_name, name, cc_toolchain_target, use_legacy_platform_definition)
+    _create_platform(ctx, image_name, name, cc_toolchain_target, use_legacy_platform_definition, exec_properties)
 
 # Creates a BUILD file with the cc_toolchain and platform targets
-def _create_platform(ctx, image_name, name, cc_toolchain_target, use_legacy_platform_definition):
+def _create_platform(ctx, image_name, name, cc_toolchain_target, use_legacy_platform_definition, exec_properties):
```
nlopezgi: alphabetical order for args

rules/rbe_repo/build_gen.bzl
```diff
@@ -132,24 +139,38 @@ def create_alias_platform(ctx, toolchain_config_spec_name, image_name, name, use
...
     template = ctx.path(Label("@bazel_toolchains//rules/rbe_repo:BUILD.platform_legacy.tpl")) if use_legacy_platform_definition else ctx.path(Label("@bazel_toolchains//rules/rbe_repo:BUILD.platform.tpl"))
     exec_compatible_with = ("\"" +
                             ("\",\n        \"").join(ctx.attr.exec_compatible_with) +
                             "\",")
     target_compatible_with = ("\"" +
                               ("\",\n        \"").join(ctx.attr.target_compatible_with) +
                               "\",")
+
+    if use_legacy_platform_definition == True and exec_properties:
```
nlopezgi: you should fail much earlier than here (around https://github.com/bazelbuild/bazel-toolchains/blob/master/rules/rbe_repo.bzl#L956)

rules/rbe_repo/build_gen.bzl
```diff
@@ -132,24 +139,38 @@ def create_alias_platform(ctx, toolchain_config_spec_name, image_name, name, use
...
+        fail("exec_properties must not be set when use_legacy_platform_definition is True")
+
+    platform_exec_properties = create_exec_properties_dict(
+        container_image = "docker://%s" % image_name,
+        os_family = "Linux",
+    )
+    if use_legacy_platform_definition == False and exec_properties:
```
nlopezgi: same, lets try to fail in the macro around https://github.com/bazelbuild/bazel-toolchains/blob/master/rules/rbe_repo.bzl#L956

WORKSPACE
eytankidron: load("//rules/experimental/rbe:exec_properties.bzl", "merge_dicts")

rules/rbe_repo.bzl
eytankidron: Yes. I call fail(...).
I prefer to do that so that we won't have to deal with breaking people who choose to use it anyway.
If we find that people want to specify container_image using this mechanism, we can open it up in the future.

rules/rbe_repo.bzl
eytankidron: Yes.

rules/rbe_repo.bzl
eytankidron: I was actually debating whether to write container-image, container_image or container image.
container-image is the name of the key in the dict.
container_image is the parameter of create_exec_properties_dict.
To avoid confusion, I think I prefer to stick with container image.

rules/rbe_repo.bzl
eytankidron: Out of curiosity, what *is* the maximum line length?
It doesn't seem to be the "standard" 80 or 100.

nlopezgi: I think its 80, but we dont enforce it via any tools, so as long as it looks decent in github side by side view it's good to go

eytankidron: done

rules/rbe_repo/build_gen.bzl
eytankidron: done

rules/rbe_repo/build_gen.bzl
eytankidron: done

rules/rbe_repo/build_gen.bzl
eytankidron: done

rules/rbe_repo/build_gen.bzl
eytankidron: done

rules/rbe_repo/build_gen.bzl
eytankidron: done

rules/rbe_repo/build_gen.bzl
eytankidron: done

PR #748, opened 2019-10-21
Add labels parameter to create_exec_properties_dict.

eytankidron:
This adds properties of the form {"label:" + label_name: label_value}


eytankidron:
> can we add a test / docs that exemplify how this new "labels" should be used. I'm not sure I understand how/why this is needed.

As with any parameter, I prefer not to document its usage here, but rather in https://cloud.google.com/remote-build-execution/docs/remote-execution-properties. Where it will be documented soon (see cl/275532252) which is still under review. I understand if you would prefer to wait for that change to land before merging this one, but I don't think the order is that important.

And as for tests, we do not have integration tests for all parameters. In particular, we don't have integration tests for the targeting parameters (such as gce_machine_type and pool) of which labels is one. Do you think it makes sense to add such a test?

In terms of explaining how these parameters work:
RBE accepts zero or more properties that looks like this: key = "labels:label_name" value = "label_value".
This means that the relevant action can *only* run on workers whose worker pool has a label "label_name" with a value "label_value".
These labels are user defined and can be associated to a worker pool using the admin API.

eytankidron:
Following our offline discussion, I added some unit tests for testing create_exec_properties_dict itself.
To do that, I defined a test rule compare_dicts_test which receives an expected dict and an actual dict and compares them.

tests/rules/exec_properties/compare.bzl
```diff
@@ -0,0 +1,49 @@
+def _test_succeeds():
```
nlopezgi: Add license header and module docstring.

tests/rules/exec_properties/compare.bzl
```diff
@@ -0,0 +1,49 @@
...
+compare_dicts_test = rule(
+    implementation = _compare_dicts_impl,
+    attrs = {
+        "expected": attr.string_dict(
+            mandatory=True,
+            doc="The expected dictionary",
+        ),
+        "actual": attr.string_dict(
+            mandatory=True,
+            doc="The expected dictionary",
```
nlopezgi: actual

eytankidron: done

tests/rules/exec_properties/compare.bzl
eytankidron: done

nlopezgi:
buildifier seems to still be complaining

PR #755, opened 2019-10-23
Add documentation for using exec_properties in rbe_autoconfig

eytankidron:


rules/rbe_repo.bzl
```diff
@@ -237,6 +237,62 @@ rbe_autoconfig(
     toolchain_config_suite_spec = your_toolchain_config_suite_spec_struct,
 )
 
+As of bazel 0.29.0, platforms support exec_properties instead of the deprecated
+remote_execution_properties to configure remote execution properties. The new
+field is a string->string dictionary rather than a proto serialized as a
+string.
+
+rbe_autoconfig now has a field use_legacy_platform_definition, which for
```
nlopezgi: Should we make the non legacy the default (this does not block this PR)?

rules/rbe_repo.bzl
```diff
@@ -237,6 +237,62 @@ rbe_autoconfig(
...
+
+load("@bazel_toolchains//rules/exec_properties:exec_properties.bzl", "create_exec_properties_dict")
+
+rbe_autoconfig(
+    name = "rbe_default",
+    use_legacy_platform_definition = False,
+    exec_properties = create_exec_properties_dict(docker_shm_size = "128m"),
+)
+
+Note the use of create_exec_properties_dict. This is a bazel macro that makes
```
nlopezgi: capitalize Bazel

rules/rbe_repo.bzl
eytankidron: We can't. Doing that would break existing users who create a platform using rbe_autoconfig and then inherit from it, and override the remote_execution_properties field.

rules/rbe_repo.bzl
eytankidron: done.

PR #763, opened 2019-11-08
Replace create_exec_properties_dict with create_rbe_exec_properties_dict

eytankidron:
And also replace merge_dicts with dicts.add

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 # based on the name of the directory bazel is cloned into.
 /bazel-*
 /tests/rbe_repo/examples/bazel_toolchains_client/bazel-*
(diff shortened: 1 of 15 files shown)


eytankidron:
> Looking in further detail. It seems like this change makes it so that all bazel-toolchains customers who just use rbe_autoconfig will need to now import skylib via repositories.bzl.
> I think we should keep using our version of merge instead of adding this dependency just for this functionality to avoid having all users need to add the call to repositories.bzl.

Summarizing our offline discussion: we will still deprecate merge_dicts() and asks users to use dicts.add() instead. However, in order not to break customers who are just using rbe_config and not calling repositories() from their WORKSPACE, we will have build_gen.bzl *not* use dicts.add() and instead use its own local version of _merge_dicts().

PR #803, opened 2020-01-13
Add ISA property

jamespthomas:
This allows for a standard instruction set architecture to bet set when using create_rbe_exec_properties_dict (as per https://github.com/bazelbuild/remote-apis/blob/master/build/bazel/remote/execution/v2/platform.md)

### DIFF
--- a/rules/exec_properties/exec_properties.bzl
+++ b/rules/exec_properties/exec_properties.bzl
@@ -189,6 +189,10 @@ PARAMS = {
         key = "gceMachineType",
         verifier_fcn = _verify_string,
     ),
+    "isa": struct(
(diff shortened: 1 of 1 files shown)


eytankidron:
@jamespthomas, why do you need the ISA key in the exec_properties dict?

Currently, PARAMS is only used in create_rbe_exec_properties_dict. create_rbe_exec_properties_dict is designed to create an exec_properties dict that is consumable by RBE, and RBE does not currently support the ISA key.

jamespthomas:
@eytankidron I misread the guidance around using this, didn't realise it was RBE specific. I'll just use a dict, sorry for the noise

PR #865, opened 2020-04-21
rbe_autoconfig: added 'pull' attribute

laszlocsomor:
The new 'pull' attribute allows controlling
whether rbe_autoconfig is allowed to pull the
container from a registry.

Disabling this can be helpful while developing a
custom container that is not yet pushed to a
registry.

### DIFF
--- a/rules/rbe_repo.bzl
+++ b/rules/rbe_repo.bzl
@@ -504,18 +504,24 @@ def _rbe_autoconfig_impl(ctx):
         ctx.report_progress("validating host tools")
         docker_tool_path = validate_host(ctx)
 
-        # Pull the image using 'docker pull'
-        pull_image(ctx, docker_tool_path, image_name)
+        if ctx.attr.pull:
+            # Pull the image using 'docker pull'
+            pull_image(ctx, docker_tool_path, image_name)
 
-        # If tag is specified instead of digest, resolve it to digest in the
-        # image_name as it will be used later on in the platform targets.
-        if ctx.attr.tag:
-            result = ctx.execute([docker_tool_path, "inspect", "--format={{index .RepoDigests 0}}", image_name])
-            print_exec_results("Resolve image digest", result, fail_on_error = True)
-            image_name = result.stdout.splitlines()[0]
-            digest = image_name.split("@")[1]
-            print("Image with given tag `%s` is resolved to '%s', digest is '%s'" %
-                  (ctx.attr.tag, image_name, digest))
+            # If tag is specified instead of digest, resolve it to digest in the
+            # image_name as it will be used later on in the platform targets.
+            if ctx.attr.tag:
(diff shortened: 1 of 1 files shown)


rules/rbe_repo.bzl
```diff
@@ -573,6 +579,11 @@ def _rbe_autoconfig_impl(ctx):
 
             # Create the versions.bzl file
             if ctx.attr.create_versions:
+                if not digest:
```
smukherj1, line 582: I think it might be better to fail this earlier and have a more specific message on the problem. See my comments above.

smukherj1, line 582: I think it might be better to fail this earlier and have a more specific message on the problem. See my comments above.

rules/rbe_repo.bzl
```diff
@@ -504,18 +504,24 @@ def _rbe_autoconfig_impl(ctx):
...
+            if ctx.attr.tag:
+                # If a container was never pushed to a registry, it has no RepoDigests.
+                result = ctx.execute([docker_tool_path, "inspect", "--format={{index .RepoDigests 0}}", image_name])
+                print_exec_results("Resolve image digest", result, fail_on_error = True)
+                image_name = result.stdout.splitlines()[0]
+                digest = image_name.split("@")[1]
+                print("Image with given tag `%s` is resolved to '%s', digest is '%s'" %
+                      (ctx.attr.tag, image_name, digest))
+        else:
+            result = ctx.execute([docker_tool_path, "images", "-q", image_name])
```
smukherj1, line 522: Suggest checking if ctx.attr.create_versions is true here and fail with the message that it's not supported when pull is false.

rules/rbe_repo.bzl
```diff
@@ -504,18 +504,24 @@ def _rbe_autoconfig_impl(ctx):
...
+
+            # If tag is specified instead of digest, resolve it to digest in the
+            # image_name as it will be used later on in the platform targets.
+            if ctx.attr.tag:
+                # If a container was never pushed to a registry, it has no RepoDigests.
+                result = ctx.execute([docker_tool_path, "inspect", "--format={{index .RepoDigests 0}}", image_name])
+                print_exec_results("Resolve image digest", result, fail_on_error = True)
+                image_name = result.stdout.splitlines()[0]
+                digest = image_name.split("@")[1]
+                print("Image with given tag `%s` is resolved to '%s', digest is '%s'" %
```
smukherj1, line 519: Suggest adding a check if digest is empty string here and failing immediately.

laszlocsomor:
> A general comment. I just realized you can set up a docker private registry (see https://docs.docker.com/registry/deploying/), push the experimental image to that and use the rbe_autoconfig rule in its current form. Does that work for you?

Yes! Thank you, I didn't know about this feature. Works perfectly.

Since this PR adds complexity that we now know is unnecessary, I'll close it.

PR #877, opened 2020-05-21
Quote docker cli path to support Windows paths with spaces

sunjayBhatia:
Otherwise if Docker is in "Program Files" or another directory with spaces, bash fails to execute it with an error: `./run_and_extract.sh: line 35: C:/Program: No such file or directory`


sunjayBhatia:
cc @wrowe

sunjayBhatia:
> 😂

that is 100% my feeling, at the last step we run into this in CI b/c docker is installed in Program Files

PR #892, opened 2020-07-09
Optimize RBE Container image existence check

smukherj1:
Use a custom Go binary to use the image manifest to verify image
existence instead of using docker pull to download all the images. The
list of image monotonically increases and the test is currently timing
out because it takes more than 30mins. The new implementation takes 11s
and gives us more room for growth.


tests/rbe_repo/BUILD
```diff
@@ -867,16 +870,31 @@ sh_test(
 
 # Verifies that all containers listed in the versions.bzl file used by
 # rbe_autoconfig are available for pulling.
-# This test requires docker to be present on the path.
-sh_test(
+rbe_container_test(
     name = "rbe_container_test",
     size = "large",
```
alex1545, line 875: nit: can this test now be classified as "small"?

smukherj1: This only affects the timeout so can be left as is. "large" sets it to 30mins.

PR #837, opened 2020-03-04
Remove unused argument in print statement

prestonvanloon:
Resolves #836

This argument was not used. I'm not sure if it was useful so this PR proposes to remove it entirely.

I'd be happy to add to the print statement in this PR if anyone feels differently.

PR #904, opened 2020-08-13
Fix print format args in RBE repo checked in configs validation

sunjayBhatia:
Supercedes https://github.com/bazelbuild/bazel-toolchains/pull/837 as it seems to have gone stale

Taken suggestion from: https://github.com/bazelbuild/bazel-toolchains/pull/837#discussion_r387843190

We run into this as an issue when generating toolchains in a workspace with Linux and Windows toolchains, generating the Linux toolchains while there are others with checked in configs seems to hit this code and fail with: `not all arguments converted during string formatting`.

### DIFF
--- a/rules/rbe_repo/checked_in.bzl
+++ b/rules/rbe_repo/checked_in.bzl
@@ -196,12 +196,13 @@ def validateUseOfCheckedInConfigs(
         env = env,
         name = name,
     ):
-        print(("%s not using checked in configs; '%s' was picked/selected " +
-               "as a candidate matching config but it does not match " +
-               "the 'env = %s', 'config_repos = %s', " +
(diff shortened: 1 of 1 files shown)


sunjayBhatia:
@googlebot I signed it!

sunjayBhatia:
@googlebot I signed it!

sunjayBhatia:
@googlebot I signed it!

sunjayBhatia:
cc @wrowe

sunjayBhatia:
cc @smukherj1 when you get a chance

sunjayBhatia:
@smukherj1 when you get a chance looks like this is all green

smukherj1:
huh, wonder where the prow bot is gone. It should have added the "lgtm" and "approval" labels and our auto merger should have automatically picked it when once the CI was green

PR #913, opened 2020-09-15
Rename target_compatible_with to internal_target_compatible_with in _rbe_config

philsc:
This essentially the same patch as
9a2fbe254946db1323f8267e46ff1f887ed356a8 except that it deals with
`target_compatible_with` instead of `exec_properties`.

The PR bazelbuild/bazel#10945 adds a new `target_compatible_with`
attribute to all rules. That includes repository rules. Currently the
patch fails in `//src/test/shell/bazel:bazel_bootstrap_distfile_test`.
The error complains about:

>  Error in repository_rule: There is already a built-in attribute 'target_compatible_with' which cannot be overridden

Since the repository rule is abstracted by a macro, it should be
fairly safe to rename the attribute to
`internal_target_compatible_with`.

### DIFF
--- a/rules/rbe_repo.bzl
+++ b/rules/rbe_repo.bzl
@@ -509,7 +509,7 @@ def _rbe_autoconfig_impl(ctx):
     docker_tool_path = None
 
     # Resolve default constraints if none set
-    target_compatible_with = ctx.attr.target_compatible_with
+    target_compatible_with = ctx.attr.internal_target_compatible_with
     if not target_compatible_with:
         target_compatible_with = _TARGET_COMPAT_WITH[os_family(ctx)]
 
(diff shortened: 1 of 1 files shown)


philsc:
I tried to test this change in a few ways, but I can't seem to make bazel not throw some kind of unrelated error. So I'm hoping that buildkite will tell me if I messed something up. My apologies.

On a related note, it looks like bazel imports version 3.1.0 of this repo. So the patch that I tested within the bazel repos actually looks differently than what you see here:
```patch
diff --git a/rules/rbe_repo.bzl b/rules/rbe_repo.bzl
index a386010..faec7e6 100644
--- a/rules/rbe_repo.bzl
+++ b/rules/rbe_repo.bzl
@@ -817,7 +817,7 @@ _rbe_autoconfig = repository_rule(
         "tag": attr.string(
             doc = ("Optional. The tag of the image to pull, e.g. latest."),
         ),
-        "target_compatible_with": attr.string_list(
+        "internal_target_compatible_with": attr.string_list(
             default = _RBE_UBUNTU_TARGET_COMPAT_WITH,
             doc = ("The list of constraints that will be added to the " +
                    "toolchain in its target_compatible_with attribute. For " +
@@ -1174,7 +1174,7 @@ def rbe_autoconfig(
         registry = registry,
         repository = repository,
         tag = tag,
-        target_compatible_with = target_compatible_with,
+        internal_target_compatible_with = target_compatible_with,
         use_legacy_platform_definition = use_legacy_platform_definition,
     )
```

philsc:
This last push is a rebase.

philsc:
Now that @katre fixed Incompatible Target Skipping via bazelbuild/bazel#12378, we don't need this patch anymore.

Issue #895, opened 2020-07-11
An error occurred during the fetch of repository 'rbe_default': key "mac os x" not found in dictionary


davidlee1435:
Hey folks,

I'm currently trying to run the examples in this repository. For context, I have a remote execution server running BuildGrid on an Amazon Linux image. I am using `bazel_toolchains` version 3.3.1.

After copying the example WORKSPACE and .bazelrc, I'm receiving this error message:

```
$ bazel clean --expunge && bazel build --config=remote //...
INFO: Starting clean.
Starting local Bazel server and connecting to it...
INFO: Invocation ID: b3a3aa84-307a-4642-91f7-33d0ae5bea1a
INFO: Options provided by the client:
  Inherited 'common' options: --isatty=1 --terminal_columns=90
INFO: Found applicable config definition common:remote in file /Users/davidlee/dev/examples/java-maven/.bazelrc: --spawn_strategy=remote --strategy=Javac=remote --strategy=Closure=remote --strategy=Genrule=remote --define=EXECUTOR=remote --remote_executor=grpc://ec2-3-101-14-83.us-west-1.compute.amazonaws.com:50051 --remote_instance_name=main --announce_rc --crosstool_top=@rbe_default//cc:toolchain --host_javabase=@rbe_default//java:jdk --javabase=@rbe_default//java:jdk --host_java_toolchain=@bazel_tools//tools/jdk:toolchain_hostjdk8 --java_toolchain=@bazel_tools//tools/jdk:toolchain_hostjdk8 --extra_execution_platforms=@rbe_default//config:platform --host_platform=@rbe_default//config:platform --platforms=@rbe_default//config:platform --extra_toolchains=@rbe_default//config:cc-toolchain
INFO: Call stack for the definition of repository 'rbe_default' which is a _rbe_autoconfig (rule definition at /private/var/tmp/_bazel_davidlee/69e0bb4508c054bbdc8964a6633bb9ad/external/bazel_toolchains/rules/rbe_repo.bzl:650:19):
 - <builtin>
 - /private/var/tmp/_bazel_davidlee/69e0bb4508c054bbdc8964a6633bb9ad/external/bazel_toolchains/rules/rbe_repo.bzl:1170:5
 - /Users/davidlee/dev/examples/java-maven/WORKSPACE:41:1
[...]

PR #935, opened 2021-01-29
Allow overriding os_family in rbe_autoconfig

kevingessner:
(Rebased and making a PR on behalf of @ulfjack, who wrote this patch)

Provides a workaround for https://github.com/bazelbuild/bazel-toolchains/issues/895.

Use like this:
rbe_autoconfig(
    name = "engflow_remote_config",
    os_family = "Linux",
    java_home = "/usr/lib/jvm/11.29.3-ca-jdk11.0.2/reduced",
    use_checked_in_confs = "Force",
)

### DIFF
--- a/rules/rbe_repo.bzl
+++ b/rules/rbe_repo.bzl
@@ -823,6 +823,11 @@ _rbe_autoconfig = repository_rule(
             doc = ("Optional. The Java release version in the container. For " +
                    " example, 11. Should only be set if java_home is set."),
         ),
+        "os_family": attr.string(
+            doc = ("Optional. The os_family to generate the config for. For example, " +
+                   "Linux or Windows (Mac OS X is not supported at this time). The default is " +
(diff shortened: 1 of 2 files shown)


smukherj1:
@kevingessner FYI we're exploring the possibility of exploring rbe_autoconfig for config generation with a standalone Go CLI (see PR #936 for a proof of concept). We expect this CLI will make it easier to generate configs (mostly for our internal config generation automation as well as for new users setting up config generation for themselves).

The CLI currently supports Linux & Windows to start out but I imagine Mac OS support should be trivial to add. We're still experimenting with this tool and will share more details in the coming weeks but initial thoughts are welcome.

Right now the Go code also potentially be used as a library and we would be open to suggestions on what components of the tool should be exported to make it more pluggable. This will be especially important for potential Windows/OSX specific support that we might not initially be able to set up testing for.

PR #936, opened 2021-02-02
Adding a prototype Go CLI to generate Bazel C++/Java toolchain configs

smukherj1:
The CLI currently:
1. Accepts a Bazel version & docker image name.
2. Spins up a docker container (requires docker to be installed locally)
3. Runs Bazel inside the container to generate C++ configs.
4. Probes the container to extract JDK installation information
5. Has the ability to presents the toolchain configs in a format that can be consumed by Bazel. 
 
This tool is meant to eventually replace `rbe_autoconfig`.

Currently supported output formats:
1. Produce a tarball that could be uploaded to a remote location or extracted as is into a Github repository and imported via a http_archive definition in a different Bazel source repository.
2. Copied into an existing repository. The Go CLI will simply copy the generated configs to a local directory. The expected use case is the local directory is in a version control source repository (e.g., locally cloned Github repository) and the changes generated by the tool are then committed to the version control system.

I've locally verified the tool works when run on a Debian based system and is able to generate configs for Bazel 4.0.0 & the default RBE toolchain container at `l.gcr.io/google/rbe-ubuntu16-04:latest` which is a Ubuntu 16.04 based Clang + JDK container. 

The tool also attempts to provide support for Windows but I haven't tested this yet.

Next steps:
1. Get initial code review from other maintainers of this repo
2. Verify the produced configs are usable with RBE for Hello World C++ & Java compilation
3. Reach out to an existing rbe_autoconfig user that uses RBE on both Linux & Windows to see if the prototype works for them.


pkg/rbeconfigsgen/options.go
```diff
@@ -0,0 +1,236 @@
+// Package rbeconfigsgen contains utilities to generate C++ & Java Toolchain configs for Bazel to be
+// used to run RBE builds
+package rbeconfigsgen
+
```
eytankidron, line 4: Out of curiosity, is there a reason why this is a separate file but not a separate package?

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,847 @@
...
+    toolchain_type = "@bazel_tools//tools/cpp:toolchain_type",
+){{ end }}
+
+platform(
+    name = "platform",
+    parents = ["@local_config_platform//:host"],
+    constraint_values = [
+{{ range .ExecConstraints }}        "{{ . }}",
+{{ end }}    ],
+    remote_execution_properties = """
```
eytankidron, line 65: Can you use here `exec_properties = create_rbe_exec_properties_dict(...)`?
See https://github.com/bazelbuild/bazel-toolchains/blob/master/rules/exec_properties/README.md#create_rbe_exec_properties_dict

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,847 @@
...
+		return
+	}
+	if _, err := runCmd(d.dockerPath, "stop", "-t", "0", d.containerID); err != nil {
+		log.Printf("Failed to stop container %v of toolchain image %v but it's ok to ignore this error if config generation & extraction succeeded.", d.containerID, d.resolvedImage)
+	}
+}
+
+// copyTo copies the local file at 'src' to the container where 'dst' is the path inside the
+// container. d.workdir has no impact on this function.
+func (d *dockerRunner) copyTo(src, dst string) error {
```
eytankidron, line 292: I think calling this function `CopyToContainer` makes it more obvious that the `src` is outside of the container and the `dst` is in the container.

Similarly replace `copyFrom` with `copyFromContainer`.

eytankidron, line 292: I think calling this function `CopyToContainer` makes it more obvious that the `src` is outside of the container and the `dst` is in the container.

Similarly replace `copyFrom` with `copyFromContainer`.

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,847 @@
...
+	if _, err := r.execCmd("touch", "WORKSPACE", "BUILD.bazel"); err != nil {
+		return "", fmt.Errorf("failed to create empty build & workspace files in the container to initialize a blank Bazel repository: %w", err)
+	}
+
+	// Backup the current environment.
+	oldEnv := r.env
+	// Create a new environment for bazelisk commands used to specify the Bazel version to use to
+	// Bazelisk.
+	bazeliskEnv := []string{fmt.Sprintf("USE_BAZEL_VERSION=%s", o.BazelVersion)}
+	r.env = []string{fmt.Sprintf("USE_BAZEL_VERSION=%s", o.BazelVersion)}
```
eytankidron, line 429: bazeliskEnv?

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,847 @@
...
+	}
+
+	return env, nil
+}
+
+// genCppConfigs generates C++ configs inside the running toolchain container represented by the
+// given docker runner according to the given options. bazeliskPath is the path to the bazelisk
+// binary inside the running toolchain container.
+// The return value is the path to the C++ configs tarball copied out of the toolchain container.
+func genCppConfigs(r *dockerRunner, o *Options, bazeliskPath string) (string, error) {
```
eytankidron, line 403: nit: in some places dockerRunner is r and in others its d.
I would suggest settling on one or the other.

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,847 @@
...
+
+	// Backup the current environment.
+	oldEnv := r.env
+	// Create a new environment for bazelisk commands used to specify the Bazel version to use to
+	// Bazelisk.
+	bazeliskEnv := []string{fmt.Sprintf("USE_BAZEL_VERSION=%s", o.BazelVersion)}
+	r.env = []string{fmt.Sprintf("USE_BAZEL_VERSION=%s", o.BazelVersion)}
+	// Always restore the old env before returning.
+	defer func() {
+		r.env = oldEnv
```
eytankidron, line 432: wondering out loud: does it make sense to pass the env as a parameter to execCmd() instead of doing this dance?

cmd/rbe_configs_gen/rbe_configs_gen.go
```diff
@@ -0,0 +1,59 @@
+package main
```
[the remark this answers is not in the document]
smukherj1: Done

pkg/rbeconfigsgen/options.go
smukherj1: I wanted to isolate the internal structs & functions & execution logic of the config generation into its own file. The options are mostly publicly isolated and are probably most interesting to anyone who's trying to figure out how to call the Run option. This file also includes logic to apply certain default based on the OS and it'll be easier for the reader if they don't need to wade through internal structs/execution logic to find it.

I don't see a strong enough reason to create a separate package though because the Options are tightly coupled with the execution logic in that they only affect the behavior of the execution logic and I would like to create additional packages unless we have a really strong reason to.

pkg/rbeconfigsgen/rbeconfigsgen.go
smukherj1: As discussed offline, I'm changing this to use a dictionary but stopping short of using create_rbe_exec_... to keep dependencies simple.

pkg/rbeconfigsgen/rbeconfigsgen.go
smukherj1: Done

pkg/rbeconfigsgen/rbeconfigsgen.go
smukherj1: Done

pkg/rbeconfigsgen/rbeconfigsgen.go
smukherj1: Switched to `d`

pkg/rbeconfigsgen/rbeconfigsgen.go
smukherj1: I considered it when writing this code and decided against it.

Rationale: There are 13 calls to execCmd and only two of them need to specify the env. Making the env a mandatory argument would add unnecessary placeholders for 11 calls. Can't use variadic args for env because I'm already using it for the command. So I preferred this dance because I only need it for two calls.

pkg/rbeconfigsgen/options.go
eytankidron: That makes sense.

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,932 @@
...
+// <config root>
+// |
+//  - cc-  C++ configs as generated by Bazel's internal C++ toolchain detection logic.
+//  - config- Toolchain entrypoint target for cc_crosstool_top & the auto-generated platform target.
+//  - java- Java toolchain definition.
+func Run(o Options) error {
+	if err := processTempDir(&o); err != nil {
+		return fmt.Errorf("unable to initialize a local temporary working directory to store intermediate files: %w", err)
+	}
+	r, err := newDockerRunner(o.ToolchainContainer, o.Cleanup)
```
eytankidron, line 880: Following up on the previous comment, could you switch this too from r to d?

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,932 @@
...
+	if _, err := d.execCmd("touch", "WORKSPACE", "BUILD.bazel"); err != nil {
+		return "", fmt.Errorf("failed to create empty build & workspace files in the container to initialize a blank Bazel repository: %w", err)
+	}
+
+	// Backup the current environment.
+	oldEnv := d.env
+	// Create a new environment for bazelisk commands used to specify the Bazel version to use to
+	// Bazelisk.
+	bazeliskEnv := []string{fmt.Sprintf("USE_BAZEL_VERSION=%s", o.BazelVersion)}
+	d.env = bazeliskEnv
```
eytankidron, line 440: Between setting `d.env = bazeliskEnv` here and setting `d.env = generationEnv` a few lines further down, you are not executing any command on d. So it looks like there's no reason to set the env here.

pkg/rbeconfigsgen/options.go
```diff
@@ -0,0 +1,258 @@
...
+	// this tool will run bazel build @foo//:blah.
+	// This field can be auto-populated with a default value. See DefaultExecOptions below.
+	CppBazelCmd string
+	// CppGenEnv are the environment variables that'll be set when running the Bazel command to
+	// generate C++ configs inside the toolchain container. Only one of CppGenEnv or CppGenEnvJSON
+	// can be specified.
+	CppGenEnv map[string]string
+	// CppGenEnvJSON is a JSON file with environment variables that'll be set when running the Bazel
+	// command to generate C++ configs inside the toolchain container. This field is union'd with
+	// CppGenEnv. Only one of CppGenEnv or CppGenEnvJSON can be specified.
```
DaveGay, line 66:
```suggestion
	// command to generate C++ configs inside the toolchain container. Only one of CppGenEnv or
	// CppGenEnvJSON can be specified.
```

eytankidron (approved):

pkg/rbeconfigsgen/rbeconfigsgen.go
smukherj1: Done

pkg/rbeconfigsgen/rbeconfigsgen.go
smukherj1: Done

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,932 @@
...
+		return err
+	}
+	return nil
+}
+
+// getEnv gets the shell environment values from the toolchain container as determined by the
+// image config. Env value set or changed by running commands after starting the container aren't
+// captured by the return value of this function.
+// The return value of this function is a map from env keys to their values. If the image config,
+// specify the same env key multiple times, later values supercede earlier ones.
```
DaveGay, line 321:
```suggestion
// The return value of this function is a map from env keys to their values. If the image config
// specifies the same env key multiple times, later values supercede earlier ones.
```

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,932 @@
...
+func (d *dockerRunner) getEnv() (map[string]string, error) {
+	result := make(map[string]string)
+	o, err := runCmd(d.dockerPath, "inspect", "-f", "{{range $i, $v := .Config.Env}}{{println $v}}{{end}}", d.resolvedImage)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get inspect the docker image to get environment variables: %w", err)
```
DaveGay, line 326:
```suggestion
		return nil, fmt.Errorf("failed to inspect the docker image to get environment variables: %w", err)
```

DaveGay (commented):
Still only halfway through the last file.

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,930 @@
...
+			break
+		case tar.TypeReg:
+			if strings.HasSuffix(h.Name, "WORKSPACE") {
+				break
+			}
+			outH := &tar.Header{
+				Name:    path.Join(pathPrefix, h.Name),
+				Size:    h.Size,
+				Mode:    h.Mode,
+				ModTime: time.Unix(0, 0),
```
DaveGay, line 655: Delete, use the value from `inTar`, or the current time?  Unless you're setting it to something nonzero, this is redundant.  Ditto below.

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,930 @@
...
+			return fmt.Errorf("unable to copy C++ configs from the C++ config tarball %q to the output tarball %q: %w", oc.cppConfigsTarball, o.OutputTarball, err)
+		}
+	}
+	if o.GenJavaConfigs {
+		if err := writeGeneratedFileToTarball(oc.javaBuild, outTar); err != nil {
+			return fmt.Errorf("unable to write the BUILD file %q containing the Java toolchain definition to the output tarball %q: %w", oc.javaBuild.name, o.OutputTarball, err)
+		}
+	}
+	if err := writeGeneratedFileToTarball(oc.configBuild, outTar); err != nil {
+		return fmt.Errorf("unable to write the crostool top/platform BUILD file %q to the output tarball %q: %w", oc.configBuild.name, o.OutputTarball, err)
```
DaveGay, line 707:
```suggestion
		return fmt.Errorf("unable to write the crosstool top/platform BUILD file %q to the output tarball %q: %w", oc.configBuild.name, o.OutputTarball, err)
```

Ditto elsewhere.

pkg/rbeconfigsgen/options.go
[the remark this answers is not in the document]
smukherj1: Done

cmd/rbe_configs_gen/rbe_configs_gen.go
```diff
@@ -0,0 +1,65 @@
...
+	outputSrcRoot    = flag.String("output_src_root", "", "(Optional) Path to root directory of Bazel repository where generated configs should be copied to. Configs aren't copied if this is blank. Use '.' to specify the current directory.")
+	outputConfigPath = flag.String("output_config_path", "", "(Optional) Path relative to what was specified to --output_src_root where configs will be extracted. Defaults to root if unspecified. --output_src_root is mandatory if this argument is specified.")
+	outputManifest   = flag.String("output_manifest", "", "(Optional) Generate a text file with details about the generated configs.")
+
+	// Optional input arguments that affect config generation for either C++ or Java configs.
+	genCppConfigs  = flag.Bool("generate_cpp_configs", true, "(Optional) Generate C++ configs. Defaults to true.")
+	cppEnvJSON     = flag.String("cpp_env_json", "", "(Optional) JSON file containing a str -> str dict of environment variables to be set when generating C++ configs inside the toolchain container. This replaces any exec OS specific defauls that would usually be applied.")
```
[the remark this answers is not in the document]
smukherj1: Done

cmd/rbe_configs_gen/rbe_configs_gen.go
```diff
@@ -0,0 +1,65 @@
...
+	// Mandatory input arguments.
+	toolchainContainer = flag.String("toolchain_container", "", "Repository path to toolchain image to generate configs for. E.g., l.gcr.io/google/rbe-ubuntu16-04:latest")
+	execOS             = flag.String("exec_os", "", "The OS (linux|windows) of the toolchain container image a.k.a, the execution platform in Bazel.")
+	targetOS           = flag.String("target_os", "", "The OS (linux|windows) artifacts built will target a.k.a, the target platform in Bazel.")
+
+	// Optional input arguments.
+	bazelVersion = flag.String("bazel_version", "", "(Optional) Bazel release version to generate configs for. E.g., 4.0.0. If unspecified, the latest available Bazel release is picked.")
+
+	// Arguments affecting output generation not specific to either C++ or Java Configs.
+	outputTarball    = flag.String("output_tarball", "", "(Optional) Path where a tarball with the generated configs will be dumped.")
```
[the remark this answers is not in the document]
smukherj1: Done

pkg/rbeconfigsgen/options.go
```diff
@@ -0,0 +1,258 @@
...
+	OSLinux = "linux"
+	// OSWindows represents Windows when selecting platforms.
+	OSWindows = "windows"
+)
+
+var (
+	validOS = []string{
+		OSLinux,
+		OSWindows,
+	}
```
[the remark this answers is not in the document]
smukherj1: I'm going to stick with the string list because:
1. I don't expect to support more than 3-4 OS so the O(n) shouldn't be a concern and is much simpler. Usually for very small lists, an array performs much better (caches, branch prediction, etc) compared to hashes.
2. Given the key join becomes more complicated, I don't see a huge value in refactoring to get rid of the strListContains

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,932 @@
...
+// OS where the OS refers to the OS of the toolchain container.
+func workdir(os string) string {
+	switch os {
+	case OSLinux:
+		return "/workdir"
+	case OSWindows:
+		return "C:/workdir"
+	}
+	log.Fatalf("Invalid OS: %q", os)
+	return "<invalid os>"
```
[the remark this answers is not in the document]
smukherj1: Switched to blank string. Thoughts on why you felt "[INVALID_OS]" would be better than "<invalid os>"?

pkg/rbeconfigsgen/rbeconfigsgen.go
smukherj1: Done

pkg/rbeconfigsgen/rbeconfigsgen.go
smukherj1: Done

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,932 @@
...
+
+// runCmd runs an arbitrary command in a shell, logs the exact command that was run and returns
+// the generated stdout/stderr. If the command fails, the stdout/stderr is always logged.
+func runCmd(cmd string, args ...string) (string, error) {
+	cmdStr := ""
+	if len(args) == 0 {
+		cmdStr = fmt.Sprintf("'%s'", cmd)
+	} else {
+		cmdStr = fmt.Sprintf("'%s %s'", cmd, strings.Join(args, " "))
+	}
```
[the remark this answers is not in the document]
smukherj1: Hmm I get `invalid operation: operator + not defined for ([]string literal) (value of type []string)compiler`. I got `strings.Join(append([]string{cmd}, args...), " ")` to work

pkg/rbeconfigsgen/rbeconfigsgen.go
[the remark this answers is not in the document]
DaveGay: It's a better marker than an empty string, if tracing/debugging, however an empty string is fine, and more in keeping with Golang's preference to put errors in an `error` return, too.

pkg/rbeconfigsgen/rbeconfigsgen.go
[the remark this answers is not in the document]
smukherj1: But nice suggestion!

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -0,0 +1,930 @@
...
+	out, err := d.execCmd(javaBin, "-XshowSettings:properties", "-version")
+	if err != nil {
+		return generatedFile{}, fmt.Errorf("unable to determine the Java version installed in the toolchain container: %w", err)
+	}
+	javaVersion := ""
+	for _, line := range strings.Split(out, "\n") {
+		// We're looking for a line that looks like `java.version = <version>` and we want to
+		// extract <version>.
+		line = strings.TrimSpace(line)
+		if !strings.HasPrefix(line, "java.version") {
```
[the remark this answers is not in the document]
smukherj1: Done

pkg/rbeconfigsgen/rbeconfigsgen.go
smukherj1: Done but I didn't get what you meant by `Ditto below`. The one below constructs a header from scratch and there's no input tarball to initialize a header from.

pkg/rbeconfigsgen/rbeconfigsgen.go
smukherj1: Done

pkg/rbeconfigsgen/rbeconfigsgen.go
DaveGay: I meant the `time.Unix(0, 0)`, in particular (sorry, still getting used to github reviews, meant to focus on just that bit, not the whole header).  Any reason NOT to use `time.Now()` in both places?

DaveGay (approved):
LGTM

pkg/rbeconfigsgen/rbeconfigsgen.go
smukherj1: Ah I see. As I commented in the code here, I noticed some weirdness with the mod times of the input tarball so I just decided to use epoch for everything to avoid the weirdness and also make it deterministic. So going to keep this for now.

PR #942, opened 2021-02-12
Adding Cloud monitoring support & PR presubmit workflow for config generation tools

smukherj1:


cmd/rbe_configs_upload/rbe_configs_upload.go
```diff
@@ -158,9 +163,48 @@ func main() {
...
+
+	result := true
+	if err := uploadConfigs(ctx); err != nil {
+		log.Printf("Configs upload failed: %v", err)
+		result = false
+	} else {
+		log.Printf("Configs uploaded successfully.")
+	}
+
+	if err := mc.ReportToolchainConfigsUpload(ctx, *monitoringDockerImage, result); err != nil {
```
eytankidron, line 204: Check if mc is nil

smukherj1: mc will never be nil in upload or e2e test

cmd/rbe_configs_gen/rbe_configs_gen.go
```diff
@@ -46,10 +50,97 @@ var (
...
+	}
+	if !(*cleanup) {
+		log.Printf("--cleanup=%v \\", *cleanup)
+	}
+	if *enableMonitoring {
+		log.Printf("--enable_monitoring=%v \\", *enableMonitoring)
+	}
+	if len(*monitoringProjectID) != 0 {
+		log.Printf("--monitoring_project_id=%q", *monitoringProjectID)
+	}
```
DaveGay, line 100: Entry for --monitoring_docker_image ?

cmd/rbe_configs_upload/rbe_configs_upload.go
```diff
@@ -91,7 +94,7 @@ func newStorage(ctx context.Context) (*storageClient, error) {
 	}
 	return &storageClient{
 		client:     c,
-		bucketName: "rbe-bazel-toolchains",
+		bucketName: "rbe-toolchain",
```
DaveGay, line 97: Will clients be looking in an old bucket?  If so, maybe delete the contents once we're rolling on the new one, so any breakages are immediate/intentional?

pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,285 @@
...
+)
+
+const (
+	// Metric types for metrics reported by this package to Cloud Monitoring. See docs for "type" in
+	// https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors#resource:-metricdescriptor.
+
+	// BEGIN Metrics for Toolchain Configs Generation
+	//
+	// All toolchain config generation metrics have the following characterics:
+	// 1. Cummulative integer tracking the number of times a toolchain configs release step
```
DaveGay, line 40: [micronit!]

```suggestion
	// 1. Cumulative integer tracking the number of times a toolchain configs release step
```

pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,285 @@
...
+	//    (generation, upload & test) runs to completion.
+	// 2. Each metric includes the following labels:
+	//    a. docker_image- A string representing the OS name of the toolchain docker image. e.g.,
+	//                     "rbe-ubuntu1604".
+	//    b. result- Bool set to true if the step succeeded.
```
DaveGay, line 45: [readability]  How about `success` or `successful` for this (instead of `result`)?

pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,285 @@
...
+	//
+	// mtypeToolchainConfigsGenRuns tracks successful runs of rbe_configs_gen i.e., configs
+	// generation.
+	mtypeToolchainConfigsGenRuns = "custom.googleapis.com/toolchain_configs/generation/runs"
```
DaveGay, line 49:
```suggestion
	mtypeToolchainConfigsGenRuns = "custom.googleapis.com/rbe_toolchain_configs/generation/runs"
```

RBE is a relatively unique global namespace term, compared to "toolchain".  Ditto for the other metrics.

pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	return nil
+}
+
+// reportCummulativeCount reports the given metric type which is expected to be of kind
+// cummulative (https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors#metrickind)
+// adding "1" to the cummulative count. Other arguments:
+// imageName: The toolchain container image for which the metric is being reported.
+// result: Indicates if the workflow was successful.
+func (c *Client) reportCummulativeCount(ctx context.Context, metricType, imageName string, result bool) error {
```
DaveGay, line 168:
```suggestion
func (c *Client) reportCumulativeCount(ctx context.Context, metricType, imageName string, result bool) error {
```

pkg/monitoring/monitoring.go
```diff
@@ -0,0 +1,285 @@
...
+	reset := &timestamp.Timestamp{
+		Seconds: c.resetTs.Unix(),
+	}
+	// For cummulative metrics, end time should be > than reset time. Thus, sleep for 2 seconds if
+	// we find less than 1s has passed since resetTs. Sleep 2s instead of 1s to avoid flakes
+	// from floating point addition errors.
+	if time.Now().Sub(c.resetTs).Seconds() < 1 {
```
DaveGay, line 175: Hmm, this Won't Happen, but if it does (I'm looking at you, Daylight Saving Time!), I'd rather there not be a lurking horror: we should either change c.resetTs before the sleep (infeasible?) or print something and check in a loop.

tests/scripts/configs_e2e/configs_e2e.go
```diff
@@ -352,8 +355,51 @@ func runTestBuild(ctx context.Context, workingDir, bazelVersion string) error {
...
+// report results to monitoring before triggering a fatal exit.
+func runTest(ctx context.Context) error {
+	m, err := downloadManifest(*manifestURL)
+	if err != nil {
+		return fmt.Errorf("unable to download the manifest from %q: %w", *manifestURL, err)
+	}
+	log.Printf("Successfully downloaded the JSON manifest from %s", *manifestURL)
+
+	if err := verifyConfigSHA(m, *configsURL); err != nil {
+		return fmt.Errorf("failed to cross-reference configs digest specified in the manifest with the configs tarball: %w", err)
```
DaveGay, line 382:
```suggestion
		return fmt.Errorf("failed to cross-check configs digest specified in the manifest with the configs tarball: %w", err)
```

cmd/rbe_configs_gen/rbe_configs_gen.go
smukherj1: Done

cmd/rbe_configs_upload/rbe_configs_upload.go
smukherj1: TL; DR no special action necessary here

No one should be using the old bucket yet because usage instructions haven't been communicated yet. I had to rename the bucket because internally the new bucket is hosted in a different GCP project and it's not trivial to move a bucket to a different project ID (can require waiting 90 days I think before a bucket ID name is available after deletion).

pkg/monitoring/monitoring.go
smukherj1: Done

pkg/monitoring/monitoring.go
smukherj1: Done

pkg/monitoring/monitoring.go
smukherj1: Done

tests/scripts/configs_e2e/configs_e2e.go
smukherj1: Done

pkg/monitoring/monitoring.go
smukherj1: Done. Replaced with "rbe/bazel-toolchains" because these configs are specific to Bazel as well

pkg/monitoring/monitoring.go
smukherj1: This is a very good point however I think I'm going to leave this as is & add a comment explaining this case isn't handled.

Rationale: If resetTs < current time, the reporting the metric will fail but the rest of the binary should be unaffected. Also, we only expect to run this tool hourly during business hours so the DST cutoff shouldn't affect our automation. Also, even if it does, at worst an alert will fire *once a year* (only when the time jumps back) and self recover within 3-4 hours (depends on the sliding window we use to calculate alert thresholds) so I don't think it's worth engineering a solution for this very rare scenario and keep the code simple instead.

DaveGay (approved):

PR #954, opened 2021-02-25
Make rbe_configs_gen more windows friendly.

rubensf:
These non-exhaustive fixes address some of the more obvious problems:
1. "find" works differently on windows.
2. '/' -> '\'
3. There may be no symlinks on the local configs.

### DIFF
--- a/pkg/rbeconfigsgen/rbeconfigsgen.go
+++ b/pkg/rbeconfigsgen/rbeconfigsgen.go
@@ -29,6 +29,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"text/template"
@@ -496,12 +497,29 @@ func genCppConfigs(d *dockerRunner, o *Options, bazeliskPath string) (string, er
 	// 2. Harden each link.
 	// 3. Archive the contents of the config output directory into a tarball.
 	// 4. Copy the tarball from the container to the local temp directory.
-	out, err := d.execCmd("find", cppConfigDir, "-type", "l")
+	var out string
+	if o.ExecOS == "windows" {
+		out, err = d.execCmd("cmd", "/r", "dir", filepath.Clean(cppConfigDir), "/a:l", "/b")
+	} else {
+		out, err = d.execCmd("find", cppConfigDir, "-type", "l")
+	}
 	if err != nil {
-		return "", fmt.Errorf("unable to list symlinks in the C++ config generation build output directory: %w", err)
+		errMsg := fmt.Sprintf("unable to list symlinks in the C++ config generation build output directory: ")
+		// Windows `dir` has a non-zero exit status if no files are found.
+		// Linux just doesn't return any files but has a zero exit.
+		switch o.ExecOS {
+		case "windows":
+			out = ""
+			log.Printf("Ignoring error indicating no symlinks were found in the Bazel output directory: %v", err)
+		default:
+			return "", fmt.Errorf("%s%w", errMsg, err)
+		}
 	}
 	symlinks := strings.Split(out, "\n")
 	for _, s := range symlinks {
+		if s == "" {
+			continue
+		}
 		resolvedPath, err := d.execCmd("readlink", s)
 		if err != nil {
 			return "", fmt.Errorf("unable to determine what the symlink %q in %q in the toolchain container points to: %w", s, cppConfigDir, err)


pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -496,25 +496,31 @@ func genCppConfigs(d *dockerRunner, o *Options, bazeliskPath string) (string, er
...
-	for _, s := range symlinks {
-		resolvedPath, err := d.execCmd("readlink", s)
-		if err != nil {
-			return "", fmt.Errorf("unable to determine what the symlink %q in %q in the toolchain container points to: %w", s, cppConfigDir, err)
-		}
-		if _, err := d.execCmd("ln", "-f", resolvedPath, s); err != nil {
-			return "", fmt.Errorf("failed to harden symlink %q in %q pointing to %q: %w", s, cppConfigDir, resolvedPath, err)
+	if err != nil {
+		log.Printf("WARNING: unable to list symlinks in the C++ config generation build output directory: %v", err)
+	} else {
```
smukherj1, line 507: Don't think this else is necessary. Set `out` to blank string if the command to list symlinks fails and you'll have an empty slice in `symlinks` and the for loop will never be entered. Avoids nesting the control flow

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -496,25 +496,31 @@ func genCppConfigs(d *dockerRunner, o *Options, bazeliskPath string) (string, er
...
 	}
-	symlinks := strings.Split(out, "\n")
-	for _, s := range symlinks {
-		resolvedPath, err := d.execCmd("readlink", s)
-		if err != nil {
-			return "", fmt.Errorf("unable to determine what the symlink %q in %q in the toolchain container points to: %w", s, cppConfigDir, err)
-		}
-		if _, err := d.execCmd("ln", "-f", resolvedPath, s); err != nil {
-			return "", fmt.Errorf("failed to harden symlink %q in %q pointing to %q: %w", s, cppConfigDir, resolvedPath, err)
+	if err != nil {
```
smukherj1, line 505: Can you check if find returns an error if symlinks aren't found? If the command behaves similar on both platforms I suggest removing `WARNING` here and just saying "no symlinks were found in the Bazel output directory that needed hardening. Ignoring the error returned by the symlinks find command".

If the error is specific to windows, then we should only ignore the error on windows.

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -379,7 +379,7 @@ func installBazelisk(d *dockerRunner, downloadDir, execOS string) (string, error
 	}
 	defer resp.Body.Close()
 
-	localPath := path.Join(downloadDir, filename)
+	localPath := filepath.Join(downloadDir, filename)
```
smukherj1, line 382: Curious: What errors do you see when using `path.Join` on Windows?

rubensf: Nvm - `PS > cmd /r /unix/style/path` indeed doesn't work. I get errors eg `Invalid Switch - "Users"`. 

Running it directly from `cmd` seems to work, but IIUC the container could be using either cmd or PS.

I tweaked it so it only cleans uses windows filepath on the `dir` command so it doesn't potentially break eg bazel paths.

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -496,12 +497,29 @@ func genCppConfigs(d *dockerRunner, o *Options, bazeliskPath string) (string, er
...
+	}
 	if err != nil {
-		return "", fmt.Errorf("unable to list symlinks in the C++ config generation build output directory: %w", err)
+		errMsg := fmt.Sprintf("unable to list symlinks in the C++ config generation build output directory: ")
+		// Windows `dir` has a non-zero exit status if no files are found.
+		// Linux just doesn't return any files but has a zero exit.
+		switch o.ExecOS {
+		case "windows":
+			out = ""
+			log.Printf("WARNING: %s%v", errMsg, err)
```
smukherj1, line 513: Not a warning because it's not really actionable. Change to `Ignoring error indicating no symlinks were found in the Bazel output directory: %v`

No need to include errmsg because it would be confusing in the logs

rubensf: Done.

smukherj1 (approved):

PR #955, opened 2021-03-02
Allow changing the CPP Toolchain.

rubensf:
The cpp default toolchain on windows was hardset to cc-compiler-k8, which
is linux exclusive. Additionally it's quite common to need to change the
compiler being used.

The drawback of this approach is that it potentially creates many repetitive
equal cc_toolchains if you use the same container with multiple toolings installed.

Long term we could add multiple cpp toolchains to configs/, but so far the limited
amount of repetition isn't harmful.


pkg/rbeconfigsgen/options.go
```diff
@@ -79,6 +79,9 @@ type Options struct {
 	// command to generate C++ configs inside the toolchain container. Only one of CppGenEnv or
 	// CppGenEnvJSON can be specified.
 	CppGenEnvJSON string
+	// CPPToolchainTarget is the toolchain to be used by the cpp configs. It overrides the default
```
smukherj1, line 82: Also, I suggest creating a new variable called `CPPToolchainTarget` at the top level in the Options struct to align with the fact that all command line variables translate to Option fields.

This way, `CppToolchainTarget` inside `PlatformParams` is only populated once we know the full target path based on the config dir (if specified)

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -651,16 +651,18 @@ func processTempDir(o *Options) error {
...
 	// Populate the C++ toolchain target if C++ config generation is enabled.
 	if o.GenCPPConfigs {
-		o.PlatformParams.CppToolchainTarget = "//cc:cc-compiler-k8"
+		if o.CPPToolchainTarget != "" {
+			o.PlatformParams.CppToolchainTarget = fmt.Sprintf("//cc:%s", o.CPPToolchainTarget)
+		}
 		if o.OutputConfigPath != "" {
-			o.PlatformParams.CppToolchainTarget = fmt.Sprintf("//%s/cc:cc-compiler-k8", path.Clean(o.OutputConfigPath))
+			// The Toolchain target should always start with `//`.
+			o.PlatformParams.CppToolchainTarget = strings.TrimPrefix(o.PlatformParams.CppToolchainTarget, "/")
```
smukherj1, line 661: This trimming should be unnecessary.

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -651,16 +651,18 @@ func processTempDir(o *Options) error {
...
+			o.PlatformParams.CppToolchainTarget = fmt.Sprintf("//%s%s", path.Clean(o.OutputConfigPath), o.PlatformParams.CppToolchainTarget)
 		}
 	} else {
+		o.PlatformParams.CppToolchainTarget = ""
```
smukherj1, line 665: I think the cpp toolchain target path construction logic has become complicated enough that we should refactor it into its own function and add unit tests. Specifically I'm interested in making sure the target path is constructed correctly for various values of `o.OutputConfigPath`

rubensf: Done.

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -648,19 +648,21 @@ func processTempDir(o *Options) error {
 	return nil
 }
 
+func genCppToolchainTarget(o *Options) string {
+	if o.OutputConfigPath != "" {
+		return fmt.Sprintf("//%s/cc:%s", path.Clean(o.OutputConfigPath), o.CPPToolchainTargetBasename)
```
smukherj1, line 653: Add a string replace for "\\" -> "/" in o.OutputConfigPath and unit test for this case. I don't think Bazel allows "\\" in labels. Also, ensure you don't mutate `o.OutputConfigPath`, so something like:

```
configPath := string.Replace(o.OutputConfigPath, "\\", "/")
if configPath != "" {
...
```

pkg/rbeconfigsgen/options.go
```diff
@@ -79,6 +79,8 @@ type Options struct {
 	// command to generate C++ configs inside the toolchain container. Only one of CppGenEnv or
 	// CppGenEnvJSON can be specified.
 	CppGenEnvJSON string
+	// CPPToolchainTarget is the toolchain to be used by the cpp configs.
+	CPPToolchainTargetBasename string
```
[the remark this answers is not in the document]
rubensf: I'm afraid that it's confusing with the `CppToolchainTarget` inside of `PlatformParams`. They technically represent the entire thing, while this only has the "name".

Should I call it `CppToolchainName`? `...TargetName`? Wdyt?

pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -648,19 +648,21 @@ func processTempDir(o *Options) error {
...
+	}
+	return fmt.Sprintf("//cc:%s", o.CPPToolchainTargetBasename)
+}
+
 // genConfigBuild generates the contents of a BUILD file with a toolchain target pointing to the
 // C++ toolchain related rules generated by Bazel and a default platforms target.
 func genConfigBuild(o *Options) (generatedFile, error) {
-	if o.PlatformParams.CppToolchainTarget != "" {
-		return generatedFile{}, fmt.Errorf("<internal error> C++ toolchain target was already set")
-	}
```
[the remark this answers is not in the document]
rubensf: Done

pkg/rbeconfigsgen/rbeconfigsgen.go
rubensf: Done. Also added some checks taht OutputConfigPath isn't an absolute path on Validation.

smukherj1 (approved):

PR #957, opened 2021-05-04
Run latest Buildifier during presubmit

fweikert:
Buildifier 0.22.0 is no longer supported (the current version is 4.0.1), thus leading to errors on CI: https://buildkite.com/bazel/bazel-toolchains/builds/24660


fweikert:
(I forgot to mark it as "ready for review")

PR #962, opened 2021-06-02
add ability to generate toolchain config w/o docker

gorilych:


gorilych:
@googlebot I signed it!

PR #965, opened 2021-07-09
Fix extension rbe_default.json -> rbe.default.tar

glukasiknuro:
In two examples the link was incorrect, leading to 404 errors.


smukherj1 (approved):
Thanks for the correction!

PR #977, opened 2021-11-30
Add flag to force java rule bypassing the bazel version check

smukherj1:
Forces the generated java toolchain rule to use the `local_java_runtime` rule instead of the older `java_runtime` rule. Currently, the tool infers this automatically from the given Bazel version but in some cases (e.g., Bazel nightly builds), the version string is blank and the new flags gives us the ability to force the new rule.


smukherj1:
FYI @coeuvre @philwo

eytankidron (approved):

PR #1054, opened 2026-01-14
Use bazelisk's verions package to compare bazelVersion.

coeuvre:
So that it supports bazel verions like `9.0.0rc5`.


DaveGay (approved):
LGTM

PR #1070, opened 2026-05-21
Add --host_bazel_path option to rbe_configs_gen

coeuvre:
This option allows specifying a local Bazel binary on the host machine. If specified, the binary is copied into the container and used for generating the toolchain configurations.

The Bazel version is automatically detected from the host binary by running it with `--version` locally. Consequently, `--bazel_version` cannot be specified alongside `--host_bazel_path`.

Added unit tests to verify:
- Automatic version detection and parsing.
- Validation rules (mutual exclusivity of HostBazelPath with BazelPath and BazelVersion).
- Robust version comparison with panic recovery for unparseable/custom versions (e.g., "development version").

### DIFF
--- a/pkg/rbeconfigsgen/rbeconfigsgen.go
+++ b/pkg/rbeconfigsgen/rbeconfigsgen.go
@@ -570,8 +570,14 @@ func genCppConfigs(d *dockerRunner, o *Options, bazelPath string) (string, error
 	return outputTarballPath, nil
 }
 
-// Returns if bazelVersion >= targetBazelVersion.
-func isBazelVersionLessThan(bazelVersion, targetBazelVersion string) bool {
+// Returns if bazelVersion < targetBazelVersion.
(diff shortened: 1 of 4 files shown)


pkg/rbeconfigsgen/rbeconfigsgen.go
```diff
@@ -571,7 +571,13 @@ func genCppConfigs(d *dockerRunner, o *Options, bazelPath string) (string, error
 }
 
 // Returns if bazelVersion >= targetBazelVersion.
```
Wyverald: this comment seems to be the opposite of what the function is doing?

Wyverald (approved):

pkg/rbeconfigsgen/rbeconfigsgen.go
coeuvre: Updated.

